PR #140, opened 2015-02-04
Removing life and use score

bronhaim:
Signed-off-by: Yaniv Bronhaim <ybronhei@redhat.com>

### DIFF
--- a/rose/server/score.py
+++ b/rose/server/score.py
@@ -15,6 +15,7 @@ def process(players, track):
 
     for player in players.itervalues():
         if player.action == actions.LEFT:
+            # the left player should be start on x=0, and right in the middle
             if player.x > 0:
                 player.x -= 1
         elif player.action == actions.RIGHT:
@@ -28,7 +29,6 @@ def process(players, track):
     positions = set()
 
     for player in sorted_players:
-        player.score += config.move_forward
         obstacle = track.get(player.x, player.y)
         if obstacle == obstacles.CRACK:
             if player.action != actions.JUMP:
@@ -36,7 +36,8 @@ def process(players, track):
                 player.y += 1
                 player.score += config.move_backward
             else:
-                player.score += config.jump
+                player.score += config.jump + config.move_forward
+
         elif obstacle in (obstacles.TRASH,
                           obstacles.BIKE,
                           obstacles.BARRIER):
@@ -49,14 +50,14 @@ def process(players, track):
                 player.y += 1
                 player.score += config.move_backward
             else:
-                player.score += config.brake
+                player.score += config.brake + config.move_forward
         elif obstacle == obstacles.PENGUIN:
             if player.action == actions.PICKUP:
                 track.clear(player.x, player.y)
                 player.y -= 1
-                player.life += 1
-                player.score += config.move_forward
+                player.score += config.score_penguin_catch + config.move_forward
         elif obstacle == obstacles.NONE:
+            player.score += config.move_forward
             if player.action in (actions.JUMP, actions.BRAKE, actions.PICKUP):
                 player.score += config.unnecessary_action
 
@@ -79,8 +80,8 @@ def process(players, track):
             elif player.x < config.matrix_width - 1:
                 player.x += 1
 
-        print 'process_actions: name=%s lane=%d pos=%d,%d life=%d response_time=%0.6f' % (
-                player.name, player.lane, player.x, player.y, player.life,
+        print 'process_actions: name=%s lane=%d pos=%d,%d score=%d response_time=%0.6f' % (
+                player.name, player.lane, player.x, player.y, player.score,
                 player.response_time)
 
         positions.add((player.x, player.y))
--- a/rose/server/score_test.py
+++ b/rose/server/score_test.py
@@ -17,7 +17,7 @@ class SinglePlayerTest(object):
         self.player = player.Player("A", car=0, lane=0)
         self.x = self.player.x
         self.y = self.player.y
-        self.life = self.player.life
+        self.score = self.player.score
         self.track.set(self.x, self.y, self.obstacle)
 
     def process(self):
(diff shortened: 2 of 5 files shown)


rose/server/score_test.py
```diff
@@ -87,10 +87,10 @@ class TestPenguin(SinglePlayerTest):
     def test_pickup(self):
         self.player.action = actions.PICKUP
         self.process()
-        # Player move up and get more life
+        # Player move up and get more score
         assert self.player.x == self.x
         assert self.player.y == self.y - 1
-        assert self.player.life == self.life + 1
+        assert self.player.score == self.score + 1
```
nirs: Shouldn't this be self.score + config.score_penguin_catch?

rose/server/score_test.py
```diff
@@ -227,14 +227,14 @@ def test_forward(self):
         self.obstacle = obstacles.PENGUIN
         self.track.set(self.x, self.y, self.obstacle)
         self.process()
-        # Player keep position but get more life
+        # Player keep position but get more score
         assert self.player.x == self.x
         assert self.player.y == self.y
-        assert self.player.life == self.life + 1
+        assert self.player.score == self.score + 1
```
nirs: should use config, not hardcoded value. Otherwise we are testing the config instead of the logic.

rose/server/score_test.py
```diff
@@ -268,85 +268,85 @@ def test_player_in_lane_wins(self):
...
+        self.player2.score = 0
         self.player2.action = actions.PICKUP
         self.process()
         # Player 1 win because it is in lane
         assert self.player1.x == 1
         assert self.player1.y == 5
-        assert self.player1.life == 0
-        # Player 2 got more life but move back
+        assert self.player1.score == 0
+        # Player 2 got more score but move back
```
nirs: Don't we add score now in these cases?

bronhaim:
Dont take it yet! just saw that its still buggy
the car can cross to other side

rose/server/score.py
```diff
@@ -14,12 +14,19 @@ def process(players, track):
...
                 player.x -= 1
+	    else:
+		player.score -= config.move_backward
+
         elif player.action == actions.RIGHT:
-            if player.x < config.matrix_width - 1:
+            if player.x < player_start_point + config.cells_per_player - 1:
                 player.x += 1
+	    else:
+		player.score -= config.move_backward
```
nirs: Please separate this to another patch, this is not related to replacing life with score.

Are you trying to prevent player from moving to the other lane? I don't think we want this behavior.

rose/server/score.py
```diff
@@ -36,7 +42,8 @@ def process(players, track):
                 player.y += 1
                 player.score += config.move_backward
             else:
-                player.score += config.jump
+                player.score += config.jump + config.move_forward
```
nirs: Not related to replacing life with score.

rose/server/score.py
```diff
@@ -49,14 +56,14 @@ def process(players, track):
                 player.y += 1
                 player.score += config.move_backward
             else:
-                player.score += config.brake
+                player.score += config.brake + config.move_forward
```
nirs: Not related

rose/server/score.py
```diff
@@ -49,14 +56,14 @@ def process(players, track):
...
         elif obstacle == obstacles.PENGUIN:
             if player.action == actions.PICKUP:
                 track.clear(player.x, player.y)
                 player.y -= 1
-                player.life += 1
-                player.score += config.move_forward
+                player.score += config.score_penguin_catch + config.move_forward
```
nirs: Not related

rose/server/score.py
```diff
@@ -49,14 +56,14 @@ def process(players, track):
...
         elif obstacle == obstacles.NONE:
+            player.score += config.move_forward
```
nirs: Not related

rose/server/score_test.py
```diff
@@ -87,10 +87,10 @@ class TestPenguin(SinglePlayerTest):
     def test_pickup(self):
         self.player.action = actions.PICKUP
         self.process()
-        # Player move up and get more life
+        # Player move up and get more score
         assert self.player.x == self.x
         assert self.player.y == self.y - 1
-        assert self.player.life == self.life + 1
+        assert self.player.score == self.score + config.score_penguin_catch + config.move_forward
```
nirs: not related (adding both pick and move forward)

bronhaim: of course it is.. this is how it works currently

rose/server/score.py
bronhaim: I Don't agree. we have to have it.. maybe not to take points in that case. but with this formula at least you know if the user moved to the other side

rose/server/score.py
```diff
@@ -14,11 +14,13 @@ def process(players, track):
...
+        player_start_point = player.lane * config.cells_per_player
         if player.action == actions.LEFT:
-            if player.x > 0:
+            # the left player should be start on x=0, and right in the middle
+            if player.x > player_start_point:
                 player.x -= 1
         elif player.action == actions.RIGHT:
-            if player.x < config.matrix_width - 1:
+            if player.x < player_start_point + config.cells_per_player - 1:
                 player.x += 1
```
nirs: Not related to replacing life with score. Please do this in another patch and explain what are you trying to do.

rose/server/score_test.py
nirs: Looks like it works like that because you changed it in this patch. If you remove everything which is not related to replacing life with score, will this test be the same?

rose/server/score.py
bronhaim: i intend to disagree. this is the logic to have it works. without it the score means nothing

rose/server/score.py
bronhaim: wahh?! I really don't get you man

rose/server/score_test.py
bronhaim: no...

bronhaim: the commit that already in changed the way the score is counted..

PR #143, opened 2015-02-11
fix score details

bronhaim:
Instead of using penguin_catch we use move_forward.
Omit unnessecery_action_score.
Decrease score after collision.
Tests fix to fit the new score code.

Signed-off-by: Yaniv Bronhaim <ybronhei@redhat.com>


nirs:
:+1:

Issue #131, opened 2015-01-02
Fair obstacle generation


nirs:
Current code generates one random obstacle per row at a random cell. This creates unfair conditions for the two players, causing a random player to win, even if the other player code is better.

When playing with same code for each client, we should expect that both clients finish in the same or very similar result, but what we see is totally random results - even when using the new code using 3 lanes for player and considering player lane on collisions.

Possible solutions:
- generate the same sequence of obstacles for both lanes
- generate the same sequence of obstacles for both lanes but different cells
- use pre-made sequences with known quality
- use pre-made sequences randomized before use

nirs:
This is already solved using option 2 (generate the same sequence of obstacles for both lanes but different cells).

@emesika, @rollandf is current code good enough? can we close this?

nirs:
Closing old issues.

PR #146, opened 2015-07-19
modifying obstacles generator

yodem:
solving issue: Fair obstacle generation #131 

Signed-off-by: yodem yotamfromm123@gmail.com

### DIFF
--- a/rose/server/track.py
+++ b/rose/server/track.py
@@ -43,14 +43,12 @@ class Track(object):
         """
         Generates new row with obstacles
 
-        Try to create fair but random obstacle stream. Each player get the same
-        obstacles, but in different cells.
+        Each player gets the same obstacle in the same cell.
         """
+        cell = random.choice(range(0, config.cells_per_player))
         row = [obstacles.NONE] * config.matrix_width
         obstacle = obstacles.get_random_obstacle()
         for lane in range(config.max_players):
             low = lane * config.cells_per_player
-            high = low + config.cells_per_player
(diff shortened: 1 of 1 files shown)


emesika:
Don't you think that this will make the game less attractive having all lanes with the same obstacles in each second ???
Please elaborate

yodem:
I do think it will make the game less attractive but in the issue #131 Nir says that one of the solutions is 
to generate the same sequence of obstacles so this way it will be 100% fair with no posibility that the person with the better code to lose.

emesika:
Yet, I think that it can be the same sequence but in random order

yodem:
what do you mean

emesika:
If you have players p1and p2 and obstacles x, y and z then p1 may have obstacles z , x and y and p2 may have x, y and z , it will be same set of obstacles but not in the same order

nirs:
Let support all 3 - the current method, getting same obstacles in random cell, getting same obstacle in same cells, and having same obstacles in different order.

Les have an ObstacleGenerator object, with 3 implementations. We will specify the name of the class in the config, or in a command line parameter.

What do you think?

nirs:
Yodem, please explain in the commit message the motivation for the change and how the change is solving the issue. "modifying obstacle generator" is not helpful to understand your change.

emesika:
+1 for Nir's 3 methods suggestion

nirs:
Closing for now, I don't think this is solve a real issue.

Issue #113, opened 2014-12-31
Life calculations should be change when driving not in your lane


emesika:
A car that drives out of its lane will be punished X2 on obtacles and get X2 on penguins

nirs:
It is very easy to avoid obstacles, so getting x2 points for penguin *not* in your lane is not a good idea. This will make lane concept useless.

nirs:
We already have logic for checking if drive is in lane, need to check how do we use it.

PR #149, opened 2015-07-31
solution for #113 issue, life calculation

yodem:
issue - Life calculations should be change when driving not in your lane #113 solved
random driver hasn't changed
In the rose/common/config.py I added a veriable which supposed to lower the score by 20 when the player or when the car is not in his lane.
At the server/player.py code, I added a function which calculate the boundaries of the lane and returns a True value when the player not in his lane.
At the score code I used the move_extra_backwards, so it will punish the player by extra 10.

### DIFF
--- a/rose/common/config.py
+++ b/rose/common/config.py
@@ -63,5 +63,6 @@ cells_per_player = matrix_width / max_players
 
 score_move_forward = 10
 score_move_backward = -10
+score_move_forward_punish = score_move_forward / 2
 score_jump = 5
 score_brake = 4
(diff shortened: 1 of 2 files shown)


nirs:
Yodem, I think we should make a more generic change. Instead of adding special movement when you are not in your lane, punish the driver by giving half of the score or double the backward movement.

yodem:
Ok, I will work on these 12 comments in sunday :) but thank you.

bronhaim:
+1 from my side. Eli, Nir - please take a look. It's simple and clean

nirs:
This pull request seems out of date. Please close it and submit a new one.

PR #158, opened 2017-08-20
Rename examples/README to examples/README.md

rollandf:


nirs:
See #159 for example.

rollandf:
@nirs , I converted to markdown as requested

examples/README.md
```diff
@@ -77,37 +89,37 @@ with the lane owner.
 Here are the possible obstacles and the action you should take when hitting
 them:
```
nirs: We need a table for the list bellow, it is not readable in the current form.

We also have an issue with the world content, it is not readable as plain text, maybe use code block for it?

examples/README.md
```diff
@@ -73,41 +86,42 @@ with the lane owner.
 |       | car#1 |       |X|       | Car#2 |      |
 |       |       |       |X|       |       |      |
 |       |       |       |X|       |       |      |
+```
```
nirs: Looks good now.

examples/README.md
```diff
@@ -73,41 +86,42 @@ with the lane owner.
...
+- obstacles.BIKE        you must return actions.RIGHT or actions.LEFT to bypass
                         the obstacle. Any other action will cause you to
                         move backward.
                         Score: 0 for LEFT/RIGHT otherwise -10
 
-* obstacles.BARRIER     you must return actions.RIGHT or actions.LEFT to bypass
+- obstacles.BARRIER     you must return actions.RIGHT or actions.LEFT to bypass
                         the obstacle. Any other action will cause you to
                         move backward.
                         Score: 0 for LEFT/RIGHT otherwise -10
```
nirs: I think this will look better using a table:
https://help.github.com/articles/organizing-information-with-tables/

rollandf: @nirs the kids will not use a MD viewer to read this. They will use Gedit, cat or Intellij.
Making this a table in MD format will be harder to read. They are using this information a lot.

nirs: Good point - but since the kids must get the source from github, why not use the documentation on github?

We can easily publish the docs by creating a "docs" directory with some markdown files, making the docs available on RedHat-Israel.github.io/ROSE/.

PR #159, opened 2017-08-20
Convert readme to markdown

nirs:
We renamed the file recently but it was using rst syntax. Github knows
how to render it, but we should use proper syntax.


rollandf:
👍

Issue #148, opened 2015-07-29
adding same car name twice kills the client with no message in server side


bronhaim:
print in server side reasonable message when client is refused to join the game.

nirs:
:+1:

nirs:
On the client side we have very clear message:
```
$ ./rose-client ../rose-dev/zigzag.py
client connected: joining as Zig Zag
client error: Player exists: Zig Zag
client disconnected: Connection to the other side was lost in a non-clean fashion: Connection lost.
```
Logging the same error on the server side will be easy.

PR #160, opened 2017-08-20
Print error message in server when adding an existing player

rollandf:
Solves #148 
Here an example of the output :
add player: Random Driver lane: 0 car: 3
Error handling message: Player exists: Random Driver

### DIFF
--- a/rose/server/main.py
+++ b/rose/server/main.py
@@ -22,6 +22,7 @@ class Player(basic.LineReceiver):
             msg = message.parse(line)
             self.dispatch(msg)
         except error.Error as e:
+            print "Error handling message: %s" % e
             msg = message.Message('error', {'message': str(e)})
             self.sendLine(str(msg))
             self.transport.loseConnection()


rose/server/game.py
```diff
@@ -56,6 +56,7 @@ def stop(self):
 
     def add_player(self, name):
         if name in self.players:
+            print 'Failed to add player:', name, 'player already exists'
```
nirs: This is not the place to fix this - please print the exception in lineReceived, so see all errors, not only this specific error.

rose/server/game.py
```diff
@@ -56,6 +56,7 @@ def stop(self):
...
             raise error.PlayerExists(name)
```
nirs: We should make sure this error has clear text when printed, so we can do later:

    except Exception as e:
        print "Error handling message: %s" % e

rollandf:
@nirs can you review?

rose/server/game.py
```diff
@@ -56,6 +56,7 @@ def stop(self):
...
         if not self.free_cars:
             raise error.TooManyPlayers()
```
nirs: Same - we want generic way to show all errors on the server side.

rollandf:
Here an example of the server output when adding an existing player:

add player: Random Driver lane: 0 car: 3
Error handling message: Player exists: Random Driver

nirs:
:+1:

PR #163, opened 2017-08-21
Provide server address as a parameter from command line

### LINKED ISSUE #221: Provide server address as a parameter from command line
Currently the server address is specified in the driver module. This does not work well in the classroom, when students have to modify the server address many times.

After discussing this in #163, we agreed on this command line inteface:
```
./rose-client [-h|--help] [-s|--server SERVER] DRIVER
```
- If the server is not specified in the command line, fallback to driver module.
- if server is not specified anywhere, use localhost
- If server is specified both in command line and driver module, command line wins (silently)

The admin tool should be modified to use the command line interface.

Relevant docs:
- https://docs.python.org/3/library/argparse.html

rollandf:
This PR adds support for providing server address as an optional parameter from command line.

The server address in the 'Driver' code is not needed. If present it will be ignored.

When running the rose-client , the 'server-address' parameter provided will be used as FQDN/IP address of the server that the client will connect to.
If it is not provided localhost (127.0.0.1) will be used.

In addition, the URL of the game is printed in the console.

This support also enable the use of compiled drivers with any server.

Here is the help output of the command:
```
./rose-client -h
usage: rose-client [-h] [--server-address SERVER_ADDRESS] driver_file

ROSE Client

positional arguments:
  driver_file           The path to the driver python module

optional arguments:
  -h, --help            show this help message and exit
  --server-address SERVER_ADDRESS, -s SERVER_ADDRESS
                        The server address to connect to. For example:
                        '10.20.30.44' or 'my-server.com'. If not specified,
                        localhost will be used.

```

Fixes #221

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,10 @@ Start up two clients:
 
     ./rose-client mydriver.py
 
+The server FQDN/IP address can be specified that way:
+
+    ./rose-client --server-address example.com mydriver.py
+
 For driver modules, see the examples directory.
 
 
--- a/README.md
+++ b/README.md
@@ -55,9 +55,9 @@ Start up two clients:
 
     ./rose-client mydriver.py
 
-The server FQDN/IP address can be specified that way:
+The server address can be specified that way (Replace '10.20.30.44' with your server address):
 
-    ./rose-client --server-address example.com mydriver.py
+    ./rose-client -s 10.20.30.44 mydriver.py
 
 For driver modules, see the examples directory.
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -27,6 +27,10 @@ Start up to 2 clients:
 
     ./rose-client mydriver.py
 
+The server FQDN/IP address can be specified that way:
+
+    ./rose-client --server-address example.com mydriver.py
+
 For driver modules, see the examples directory.
 
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -27,9 +27,9 @@ Start up to 2 clients:
 
     ./rose-client mydriver.py
 
-The server FQDN/IP address can be specified that way:
+The server address can be specified that way (Replace '10.20.30.44' with your server address):
 
-    ./rose-client --server-address example.com mydriver.py
+    ./rose-client -s 10.20.30.44 mydriver.py
 
 For driver modules, see the examples directory.
 
--- a/rose/client/main.py
+++ b/rose/client/main.py
@@ -97,10 +97,14 @@ def load_driver_module(file_path):
 def main():
     logging.basicConfig(level=logging.INFO)
     parser = argparse.ArgumentParser(description="ROSE Client")
-    parser.add_argument("driver_file", help="The path to the driver file")
-    parser.add_argument("--server-address", dest="server_address",
-                        help="The server FQDN/IP address to connect to."
-                        " If not specified, localhost will be used.")
+    parser.add_argument("--server-address", "-s", dest="server_address",
+                        default="localhost",
+                        help="The server address to connect to."
+                             " For example: '10.20.30.44' or 'my-server.com'."
+                             " If not specified, localhost will be used.")
+    parser.add_argument("driver_file",
+                        help="The path to the driver python module")
+
     args = parser.parse_args()
 
     try:
@@ -109,14 +113,10 @@ def main():
         log.error("error loading driver module %r: %s", args.driver_file, e)
         sys.exit(2)
 
-    server_address = "127.0.0.1"
-    if args.server_address:
-        server_address = args.server_address
-
-    reactor.connectTCP(server_address, config.game_port,
+    reactor.connectTCP(args.server_address, config.game_port,
                        ClientFactory(driver_mod.driver_name, driver_mod.drive))
 
-    url = "http://%s:%d" % (server_address, config.web_port)
-    log.info("Game URL: %s" % url)
+    url = "http://%s:%d" % (args.server_address, config.web_port)
+    log.info("Please open %s to watch the game." % url)
 
     reactor.run()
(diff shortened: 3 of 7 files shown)


nirs:
Can you describe the behavior when both driver code and command line contain different server addresses?

rose/client/main.py
```diff
@@ -74,6 +74,14 @@ def main():
         d = {}
         exec f in d, d
 
-    reactor.connectTCP(d['server_address'], config.game_port,
+    if len(sys.argv) == 3:
+        server_address = sys.argv[2]
+    else:
+        server_address = d['server_address']
+
+    print ('Client \"%s\" connecting to Server \"%s\"'
```
nirs: Escaping quotes is not needed when using single quotes. Also, if you want to quote the arguments (good idea when the value provided by the user, you should use %r - which show the value quoted.

print is not a function in python 2. Using print() works since it ("string) is a string, but it confuse the reader. We don't support python 3 yet and converting to python 3 requires importing print_function from _future__, you can see how this is done in vdsm.

I would also change the message to:

   print "Client %r connecting to %r" % (server_address, d["driver_name"])\

Or:
    print "Connecting to %r as %r" %  (server_address, d["driver_name"])

rose/client/main.py
```diff
@@ -74,6 +74,14 @@ def main():
         d = {}
         exec f in d, d
 
-    reactor.connectTCP(d['server_address'], config.game_port,
+    if len(sys.argv) == 3:
+        server_address = sys.argv[2]
```
nirs: This is not a good practice for optional argument. In this case we need a proper option:

    rose-client --server-address example.com my_driver.py

rose/client/main.py
```diff
@@ -74,6 +74,14 @@ def main():
...
+    else:
+        server_address = d['server_address']
```
nirs: This will ignore the user specified server_address in the driver module. I think we should either fail or warn about this. Or maybe remove the server address from the driver file and make the server address required in the client command line, same as the admin command line.

rose/client/main.py
yoavtzelnick: I agree that argparse is probably the best solution for this.

rollandf:
@nirs please review last version. Thanks.

README.md
```diff
@@ -55,6 +55,10 @@ Start up two clients:
 
     ./rose-client mydriver.py
 
+The server FQDN/IP address can be specified that way:
+
+    ./rose-client --server-address example.com mydriver.py
+
```
nirs: The previous example show starting a server at <server-address>, so the client example should use the same server address.

Lets amend the previous example instead of adding a new one.

Kids are not expected to know what his FQDN or IP address, best just use the term "server address".  To make this clear, we should show an example like:

    ./rose-client --server-address 1.2.3.4 my_driver.py

Since most server will not have a FQDN, best show an IP address in the example.

docs/development.md
```diff
@@ -27,6 +27,10 @@ Start up to 2 clients:
 
     ./rose-client mydriver.py
 
+The server FQDN/IP address can be specified that way:
+
+    ./rose-client --server-address example.com mydriver.py
+
```
nirs: Update after modifying readme.

rose/client/main.py
```diff
@@ -95,15 +96,27 @@ def load_driver_module(file_path):
...
 def main():
     logging.basicConfig(level=logging.INFO)
-    if len(sys.argv) < 2:
-        log.info('usage: rose-client drive-module')
-        sys.exit(2)
+    parser = argparse.ArgumentParser(description="ROSE Client")
+    parser.add_argument("driver_file", help="The path to the driver file")
+    parser.add_argument("--server-address", dest="server_address",
+                        help="The server FQDN/IP address to connect to."
+                        " If not specified, localhost will be used.")
```
nirs: Please define the arguments int the same order we expect them:

    --server-address
    driver_file

Regarding server address - we like to have both short and long option. --server-address is very readable but pain to type again and again, -s will be much nicer.

We can do:

    parser.add_argument("--server-address", "-s", dest="server_address", ...)
    
If server address is not specified, we can define a default using default="localhost".

In the help, please do not use terms like FQDN or IP address, instead use the term "server address" and show example of both  "example.com" and "192.168.0.100".

In the driver_file help, write: "The path to the driver python module".

rose/client/main.py
```diff
@@ -95,15 +96,27 @@ def load_driver_module(file_path):
...
+        driver_mod = load_driver_module(args.driver_file)
     except ImportError as e:
-        log.error("error loading driver module %r: %s", sys.argv[1], e)
+        log.error("error loading driver module %r: %s", args.driver_file, e)
         sys.exit(2)
 
-    reactor.connectTCP(driver_mod.server_address, config.game_port,
+    server_address = "127.0.0.1"
+    if args.server_address:
+        server_address = args.server_address
```
nirs: No need to handle missing server_address - define a default in the parser.

rose/client/main.py
```diff
@@ -95,15 +96,27 @@ def load_driver_module(file_path):
...
+
+    reactor.connectTCP(server_address, config.game_port,
                        ClientFactory(driver_mod.driver_name, driver_mod.drive))
+
+    url = "http://%s:%d" % (server_address, config.web_port)
+    log.info("Game URL: %s" % url)
```
nirs: Maybe:

    "Please open http://... to watch the game"

We can add an option to open a browser at this address later.

docs/development.md
rollandf: Done

README.md
rollandf: Done

rose/client/main.py
rollandf: done

rose/client/main.py
rollandf: done

rose/client/main.py
rollandf: done

PR #182, opened 2017-10-11
Web: Draw finish line.

rollandf:


rose/web/game.js
```diff
@@ -332,6 +337,30 @@ var ROSE = (function() {
         }
     }
 
+    function FinishLine(loader) {
+        this.texture = null;
+        this.timeleft = 0;
```
nirs: 0 is not a good default, maybe null?

rose/web/game.js
```diff
@@ -332,6 +337,30 @@ var ROSE = (function() {
...
+        var self = this;
+        loader.load("res/end/final_flag.png", function(img) {
+            self.texture = img;
+        });
+    }
+
+    FinishLine.prototype.update = function(state) {
+        this.timeleft = state.timeleft;
```
nirs: We should protect from negative value here, since negative value will draw the finish line incorrectly.

We can use Math.max(state.timeleft, 0)

rose/web/game.js
```diff
@@ -332,6 +337,30 @@ var ROSE = (function() {
...
+    }
+
+    FinishLine.prototype.draw = function(ctx) {
+        if (this.timeleft > Config.finish_line_duration) {
+            return;
+        }
+        //Start at row 0, then move down until row finish_line_duration
```
nirs: //Start -> // Start

rose/web/game.js
rollandf: done

rose/web/game.js
rollandf: done

rose/web/game.js
rollandf: done

nirs:
Merged manually in #182

PR #185, opened 2017-10-18
Add player name label to cars in web UI

rollandf:
screenshot from 2017-10-18 16-30-26

### DIFF
--- a/rose/web/game.js
+++ b/rose/web/game.js
@@ -327,6 +327,10 @@ var ROSE = (function() {
     }
 
     Cars.prototype.draw = function(ctx) {
+        ctx.fillStyle = "rgb(0, 0, 0)";
+        ctx.textBaseline = "top";
+        ctx.font = "bold 15px sans-serif";
+        ctx.textAlign = "center";
         var i;
         for (i = 0; i < this.players.length; i++) {
             var player = this.players[i];
@@ -334,6 +338,9 @@ var ROSE = (function() {
             var x = Config.left_margin + player["x"] * Config.cell_width;
             var y = Config.dashboard_height + player["y"] * Config.row_height;
             ctx.drawImage(img, x, y);
+            var car_center = x + (img.width / 2);
+            var car_bottom = y + img.height;
+            ctx.fillText(player.name, car_center, car_bottom + 5);
         }
     }
 


rose/web/game.js
```diff
@@ -327,13 +327,19 @@ var ROSE = (function() {
...
+        ctx.font = "bold 15px sans-serif";
+        ctx.textAlign = "center";
         var i;
         for (i = 0; i < this.players.length; i++) {
             var player = this.players[i];
             var img = this.textures[player["car"]];
             var x = Config.left_margin + player["x"] * Config.cell_width;
             var y = Config.dashboard_height + player["y"] * Config.row_height;
             ctx.drawImage(img, x, y);
+            var text = player.name;
```
nirs: We can use player.name directly.

rose/web/game.js
```diff
@@ -327,13 +327,19 @@ var ROSE = (function() {
...
+            ctx.fillText(text, x + (img.width/2) , y + Config.row_height + 10);
```
nirs: I think we should use img.height for locating the label, not raw height.

I think calculating label x and y separately would be more clear.

This diff on your patch seem to work and looks good:

```
diff --git a/rose/web/game.js b/rose/web/game.js
index 93f590e..f2165ce 100644
--- a/rose/web/game.js
+++ b/rose/web/game.js
@@ -328,7 +328,7 @@ var ROSE = (function() {
 
     Cars.prototype.draw = function(ctx) {
         ctx.fillStyle = "rgb(0, 0, 0)";
-        ctx.textBaseline = "center";
+        ctx.textBaseline = "top";
         ctx.font = "bold 15px sans-serif";
         ctx.textAlign = "center";
         var i;
@@ -338,8 +338,9 @@ var ROSE = (function() {
             var x = Config.left_margin + player["x"] * Config.cell_width;
             var y = Config.dashboard_height + player["y"] * Config.row_height;
             ctx.drawImage(img, x, y);
-            var text = player.name;
-            ctx.fillText(text, x + (img.width/2) , y + Config.row_height + 10);
+            var car_center = x + (img.width / 2);
+            var car_bottom = y + img.height;
+            ctx.fillText(player.name, car_center, car_bottom + 5);
         }
     }
 
```

rose/web/game.js
```diff
@@ -327,13 +327,19 @@ var ROSE = (function() {
     }
 
     Cars.prototype.draw = function(ctx) {
+        ctx.fillStyle = "rgb(0, 0, 0)";
+        ctx.textBaseline = "center";
```
nirs: We should use "top" here - center is not valid value.
See https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/textBaseline

Issue #195, opened 2017-10-19
Update readme for new makefile


nirs:
The readme show how to run the project installed using pip, we have now pipenv support and a makefile making it much easier to use.

PragyaM:
Hi there! I'm keen to pick up this task :)

nirs:
Hi @PragyaM, you are welcome!

surajpaib:
Hey, @PragyaM. Just read your comment on the thread. I can close the issue if you'd like to send a PR.

PragyaM:
Hey @surajpaib - that's all good! I don't want to double up on work so I'll find something else to work on :)

nirs:
Playing with the makefile show that it is not the right solution, so I'm closing this issue as wontfix.

PR #197, opened 2017-10-19
Updated README for new makefile #195

surajpaib:


README.md
```diff
@@ -40,14 +44,23 @@ You can also install packages from your distribution.
...
+
+Alternatively, Start the server on some machine:
 
     ./rose-server
 
 Open a browser at http://\<server-address\>:8880 to view and control the game.
 
 Start up to 2 clients:
 
+	    make client
```
nirs: Unfortunately this does not work - the client needs a driver module argument.

I think we need to remove the makefile instead of updating the readme. It helps only with the initial setup, and using a makefile breaks compatibility on windows.

README.md
```diff
@@ -40,14 +44,23 @@ You can also install packages from your distribution.
...
+    
+Or alternatively,
```
nirs: Having alternatives for everything is confusing. We should have one simple way that works for everyone.

README.md
```diff
@@ -57,6 +70,9 @@ Command line interface
 ----------------------
 
 You can control the game from the command line using the rose-admin tool.
+	
+    make admin
```
nirs: Same issue, a server address is needed, and then a command (e.g. start) with optional arguments (e.g set-rate 5)

nirs:
@surajpaib closing this as we plan to remove the makefile, see #203. Thanks for your work!

Issue #192, opened 2017-10-19
Modernize imports


nirs:
Replace the old imports like:

````
import track
````
With modern imports:
```
from rose.server import track
```
Or:
```
from . import track
```

nirs:
@moshe742, can you post your PR for the server package?

moshe742:
Done the PR

nirs:
@moshe742 we need the same change also in the client and common packages, @andrew-jung sent #198

moshe742:
Should I start on both or only common (since @andrew-jung works on client)?

nirs:
@moshe742 andrew sent both server and client chages, but not common, so you can work on common.

moshe742:
In common there is only one import which is written as
`from rose.common import error`
Should I make the change for consistency or keep this?
That is the only place we do an import from this package so except this there is no need for a change on the imports

nirs:
@moshe742 having consistent import is nice, please post another PR for this.

nirs:
@moshe742, can complete this PR?

nirs:
@moshe742, I merged your PR, do you want to complete the client package?

moshe742:
yes, will open a PR soon

nirs:
We have one import to fix in the common package.

PR #217, opened 2017-10-20
Modernize imports of rose.server

moshe742:
for issue #192

### DIFF
--- a/rose/server/game.py
+++ b/rose/server/game.py
@@ -5,9 +5,9 @@ import os
 from twisted.internet import reactor, task
 
 from rose.common import actions, config, error, message, obstacles  # NOQA
-import track
-import player
-import score
+from . import track
+from . import player
+from . import score
 
 log = logging.getLogger('game')
 
--- a/rose/server/player_test.py
+++ b/rose/server/player_test.py
@@ -1,4 +1,4 @@
-import player
+from . import player
 
 
 def test_in_lane():
--- a/rose/server/score_test.py
+++ b/rose/server/score_test.py
@@ -1,7 +1,7 @@
 from rose.common import actions, config, obstacles
-import track
-import player
-import score
+from . import track
+from . import player
+from . import score
 import pytest
 
 


rose/server/game.py
```diff
@@ -3,9 +3,9 @@
 from twisted.internet import reactor, task
 
 from rose.common import actions, config, error, message, obstacles
-import track
-import player
-import score
+from . import track
+from . import player
+from . import score
```
nirs: Looks good

rose/server/player_test.py
```diff
@@ -1,4 +1,4 @@
-import player
+from . import player
```
nirs: Looks good

rose/server/score_test.py
```diff
@@ -1,8 +1,8 @@
 from rose.common import actions, config, obstacles
-import track
-import player
-import score
-import pytest
+from . import track
+from . import player
+from . import score
```
nirs: Looks good...

rose/server/score_test.py
```diff
@@ -1,8 +1,8 @@
...
+from . import pytest
```
nirs: But pytest is not a module this package, but a 3rd party module. This change is not needed.

nirs:
@moshe742, the tests fail with this change, see:
https://travis-ci.org/RedHat-Israel/ROSE/builds/290612341

rose/server/score_test.py
moshe742: First, how can I run the tests on my machine?
I don't have for some reason pytest on the virtualenv (or should I run some other command?)

I fixed the import and did another commit and push but not seeing it here, should I do another pull request?

nirs: To install the development packages:

    pipenv install --dev

(We need to document this in the readme)

If you fix this in the next commit, pushing again to github should show it here.

If you amended the original commit, you need to push with --force to replace the original commit on gituhub,

Your PR will be updated automatically in both ways.

nirs:
@moshe742 looks like you have bad rebase in your history, the first patch appears twice.

You can fix this by interactive rebase:

```
git checkout imports_server
git rebase -i master
```

Now delete the patch appearing twice, save and quit.

nirs:
@moshe742 can you rebase?

nirs:
@moshe742, I rebased and merge your changes in #1bf6c1bfe63ceb except the first commit which introduces pep8 violation. Thanks!

Issue #177, opened 2017-09-08
Web demo


nirs:
Have a static demo playing a pre-created (or randomly generated) game. Can be served as part of the documentation.

The game can be simulated by client side "server" receiving control requests (e.g. start/stop/rate) and emitting game update messages, using a pre-created game state.

Being able to replace a real sever with client side fake server is a good idea regardless the demo, for testing the web interface. Currently the server is accesses via the Client class and also directly via jquery.post. The code to post message should move into the Client.

PR #219, opened 2017-10-20
Remove pygame code (fixes #186)

### LINKED ISSUE #186: Remove pygame code
Since we have now a web interface, the pygame code is not needed.

The changes needed:
- remove the rendering code in rose/client/game.py
- remove the module rendering the components in rose/client/
- remove the configuration options in rose/common/config.py needed only by the client
- move modules from rose/common to rose/server that are used now only by the server
- remove pygame from the requirements.txt / Pipfile

Note: removing code without leaving unused stuff is not easy.

Ditti4:
The commits in this pull request will (hopefully) remove all the pygame-related code, including all the rendering code and hopefully leaving no unused code behind, like it was requested in #186.

If you happen to find anything that you don't agree with or that you'd like me to change, just let me know. :)


nirs:
@Ditti4, thanks for your contribution!

rose/client/finish.py
```diff
@@ -11,16 +10,5 @@ def __init__(self):
 
     # Component interface
 
-    def init(self):
-        self.texture = pygame.image.load(config.finish_line_png)
-
     def update(self, info):
         self.timeleft = info["timeleft"]
```
nirs: This is used only for rendering, I think we can remove the entire module.

rose/client/dashboard.py
```diff
@@ -18,31 +17,6 @@ def __init__(self):
         self.players = {}
         self.timeleft = config.game_duration
 
-    def init(self):
-        self.texture = pygame.image.load(config.dashboard_png)
-
     def update(self, players, timeleft):
         self.timeleft = timeleft
         self.players = players
```
nirs: The dashboard is used only for rendering, we can remove the entire module.

rose/client/game.py
```diff
@@ -29,49 +27,19 @@ def __init__(self, client, name, drive_func):
...
-        self.dashboard.init()
-        for car in self.cars:
-            car.init()
-        self.finish_line.init()
-        self.draw(self.surface)
-
     def update(self, info):
         self.track.update(info)
         self.players = {p["name"]: p for p in info['players']}
         self.dashboard.update(self.players, info["timeleft"])
```
nirs: Can be removed

rose/client/game.py
```diff
@@ -29,49 +27,19 @@ def __init__(self, client, name, drive_func):
...
         for player in self.players.itervalues():
             self.cars[player['car']].update(player)
         self.finish_line.update(info)
```
nirs: Can be removed

Ditti4:
Removing those two modules seems to have worked pretty well. I must've overlooked that those weren't used anymore. Sorry about that and thanks for pointing it out!

rose/common/config.py
```diff
@@ -9,13 +9,7 @@
 
 game_rate = 1.0
 game_duration = 60
-
-# Client
-
-frame_rate = 30
 number_of_cars = 4
-car_jitter = 10
-play_sound = True
```
nirs: The "Dashboard", "Finish line", and "Window" are also not needed now.

The the "Files" section, we can remove the paths to the various resources (e.g. obstables_dir). I think the server is using only the install_dir to serve the resources.

requirements.txt
```diff
@@ -1,5 +1,4 @@
 autobahn
-pygame
```
nirs: We introduced duplication when adding the pipenv configuration. Note to self: open an issue for generating the requirements file from pipenv.

nirs:
@Ditti4, Removing the rest of the unused configuration, it looks ready.

rose/common/config.py
Ditti4: > The the "Files" section, we can remove the paths to the various resources (e.g. obstables_dir). I think the server is using only the install_dir to serve the resources.

Yup, just tried that. Only keeping the `install_dir` (well, and `web_root` and `res_root`) seems to work. I wasn't sure about removing that first so I kept it in. Thanks!

nirs:
@Ditti4 but testing show that moving error to server package is incorrect - it is used by common/message. We need to move it back to common.

```
$ ./rose-server 
  File "./rose-server", line 3, in <module>
    from rose.server import main
  File "/home/nsoffer/src/rose/rose/server/main.py", line 9, in <module>
    from . import game, net
  File "/home/nsoffer/src/rose/rose/server/game.py", line 5, in <module>
    from rose.common import config, message, obstacles
  File "/home/nsoffer/src/rose/rose/common/message.py", line 2, in <module>
    from rose.common import error
ImportError: cannot import name error
```
```
$ ./rose-client examples/none.py 
  File "./rose-client", line 3, in <module>
    from rose.client import main
  File "/home/nsoffer/src/rose/rose/client/main.py", line 6, in <module>
    from rose.common import config, message
  File "/home/nsoffer/src/rose/rose/common/message.py", line 2, in <module>
    from rose.common import error
ImportError: cannot import name error
```

Unfortunately testing coverage is very low, so you must run both server and clients to check that everything still work.

1. start the server
2. in another shell, start a client with exampes/none.py
3. in another shell, start a client with examples/random_driver.py
4. open a browser at http://localhost:8880/
5. Start a game

Thanks!

Ditti4:
Surprisingly

Ditti4:
Hmm, the none example still uses the actions module. Should we move that back as well, considering it's not only used by the server this way either?

nirs:
@Ditti4 Right, the actions module is used by the driver code, we need to keep this module in the common package as well.

For another PR: It would be nice to write a test for client code loading a driver module, detecting such issues.

nirs:
@Ditti4 useful way to avoid stale pyc files issue is to clean your tree before testing:

    git clean -dxf

Note that it will delete anything which which git does not know about, including temporary files you may have created ;-)

Ditti4:
Do you mind if I go ahead and just remove the previous commit (which moved the two modules) using a rebase so the history gets cleaned up a little or would you prefer a revert?

nirs:
@Ditti4 I like clean history. Best remove the commit using "git rebase -i" and push again using --force. I would like to rebase and merge the commits, without squashing them.

Ditti4:
@nirs Should I also mark the two commits removing more modules and more config options as fixups of their original commits or do you want me to keep those separate?

Locally I currently have them marked as fixup so we end up with total amount of three commits.

nirs:
@Ditti4 yes, fixing the original commit instead of adding patches is preferred.

Ditti4:
Then there you go, all fixed up and hopefully final. Thanks for your guidance! :)

nirs:
@Ditti4 thanks for this fine work!

Would you like to try a more interesting and much harder task? see #177

Issue #187, opened 2017-10-19
Heroku integration


nirs:
Beni suggests in #175 that it will be easier to integrate with Heroku: 
https://devcenter.heroku.com/articles/heroku-button

The goal of this integration is allow anyone to deploy the rose-server on Heroku with few clicks.

The expected flow:
1. User clicks a deploy-on-heroku button in the project README
2. System open Heroku deploy page, with most parameters already filled up
3. User enter only the details that cannot be automated, such as application name (e.g. my-rose-server)
4. User confirm the deployment on Heroku
5. Application is deployed on Heruku, ready to accept connections from rose-client(s) on my-rose-server.herokuapp.com:8888, and showing the game UI at http://my-rose-server.herokuapp.com:8880/

Deployment details:
- Install the rose/common and rose/server packages
  - rose/client is not needed, but it is ok to install it for simplicity
- install the rose-server script
- When the application starts, rose-server should run
- When the application stops, rose-server process should terminate
- The Heroku instance should allow incoming traffic to ports 8888 (client port) and 8880 (web port)

Testing the deployment:
1. Open  http://my-rose-server.herokuapp.com:8880/ - should show an empty game
2. Create driver.py module with the server address:
```
$ cat driver.py
from rose.common import obstacles, actions

server_address = "my-rose-server.herokuapp.com"
driver_name = "My Driver"

def drive(world):
    return actions.NONE
```
3. Connect client to server:
```
./rose-client driver.py
```
4. Start the game in the web UI

The game should run, showing the connected client.

I'm not sure all this is possible, will needs some research.

mayukh18:
Taking this up.

nirs:
Thanks @mayukh18!

shiramax:
I'm Closing this issue because it's open too long and we don't have the need or the capacity to work on it.
please feel free to re-open in if you feel otherwise.

PR #222, opened 2017-10-21
Added heroku integration in readme

mayukh18:
PR for #187

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,0 +1,6 @@
+{
+  "name": "ROSE",
+  "description": "This project is a game that has been developed in order to help teach kids python. The students need to code the behavior of the car in order to achieve the best score.",
+  "repository": "https://github.com/RedHat-Israel/ROSE",
+  "keywords": ["race", "game", "python"]
+}
--- a/app.json
+++ b/app.json
@@ -1,6 +1,6 @@
 {
   "name": "ROSE",
-  "description": "This project is a game that has been developed in order to help teach kids python. The students need to code the behavior of the car in order to achieve the best score.",
+  "description": "ROSE project car race game.",
   "repository": "https://github.com/RedHat-Israel/ROSE",
   "keywords": ["race", "game", "python"]
 }
--- a/README.md
+++ b/README.md
@@ -25,6 +25,8 @@ For each type of obstacles there is a different action, and different points.
 
 See [examples/README](examples/README) for explanation on how to write a driver module.
 
+[![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy)
+
 
 ## Requirements
 


app.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "ROSE",
+  "description": "ROSE project car race game.",
+  "repository": "https://github.com/RedHat-Israel/ROSE",
+  "keywords": ["race", "game", "python"]
```
nirs: This should also describe:
- The server is started by running the ```rose-server``` script
- Port 8880 should be open to receive web clients
- Port 8888 should be open to receive rose-client(s)

See the updated issue for more details.

app.json
```diff
@@ -0,0 +1,6 @@
...
+}
```
nirs: Newline missing here.

app.json
cben: Indeed, tried the button now from your fork https://github.com/mayukh18/ROSE/ => https://cben-rose-test.herokuapp.com, says "Application Error", heroku logs show:
```
2017-10-21T17:31:29.442324+00:00 heroku[router]: at=error code=H14 desc="No web processes running" method=GET path="/" host=cben-rose-test.herokuapp.com request_id=68b48dde-386d-4756-a4bd-a5901ea49fe5 fwd="141.226.169.57" dyno= connect= service= status=503 bytes= protocol=https
```
https://cben-rose-test.herokuapp.com:8880/ doesn't work either, connection refused.

- I don't see anything in https://devcenter.heroku.com/articles/app-json-schema or elsewhere about specifying what port we listen on; we [have to respect PORT env var](https://devcenter.heroku.com/articles/dynos#local-environment-variables) that heroku provides.

- On the outside, heroku router only listens on http[s] :80 and :443.  That's fine for user interface, maybe will need to document that :8880 is not needed with heroku.

- The nastier implication is you [can't have 2nd outside port for rose-client](https://stackoverflow.com/questions/40029675/open-two-different-ports-in-heroku-app).  The client will have to connect on 80/443, and will need another mechanism for server to differentiate connection type (URL path? content-type?  user-agent (meh)?  custom header?).

mayukh18: Yes I too didn't find any documentation related to servers and ports in https://devcenter.heroku.com/articles/app-json-schema and related articles. Second thing which I know is true is that heroku pre-assigns $PORT. What we need is a **Procfile**

nirs:
Without a way to open a port for the client, Heroku is useless for us unless we redesign the way the client communicates with the server.

I think we should stop this effort for now, and try OpenShift, it may be more flexible.

mayukh18:
Yes without a configurable port it can't be used properly.

nirs:
@mayukh18 see #239, with this we should be able to deploy on Heroku.

PR #224, opened 2017-10-21
Support compiled driver modules

### LINKED ISSUE #141: Support compiled driver modules
When running a client you need to provide a driver python module:

    ./rose-client mydriver.py

This is fine for developing a driver code, but a student may like test her code against an example driver. But providing a good driver example code will spoil the challenge of developing a driver module. We like to provide pre-complied driver modules (see #137) in .pyc format, so this will work:

    ./rose-client example_driver.pyc

When loading pyc file, we can create a code object:
```
data = open('drivers/score.pyc').read()
magic_number = data[:4]
assert magic_number == imp.get_magic()
bytecode = data[8:]
code = marshal.loads(bytecode)
```

But I don't think we should use this method - this will be hard to maintain when python change the .pyc header format - this was recently changed in python 3.7.

I think the best way to solve this would be to use dynamic import using importlib.
https://docs.python.org/3/library/importlib.html

yoavtzelnick:
Fixes #141


rose/client/main.py
```diff
@@ -70,10 +72,8 @@ def main():
         print 'usage: rose-client drive-module'
         sys.exit(2)
 
-    with open(sys.argv[1]) as f:
-        d = {}
-        exec f in d, d
+    d = import_module(splitext(sys.argv[1])[0])
```
nirs: Can you separate the module name extraction to separate line to make the code easier to follow?

For example:

    module_name = splitext(sys.argv[1])[0]

This code will work only when using:

    rose-client driver.py

But we have to support also importing from another directory:

    rose-client ../my-examples/driver.py

The import fails now:

```
$ ./rose-client examples/none.py 
  File "./rose-client", line 4, in <module>
    main.main()
  File "/home/nsoffer/src/rose/rose/client/main.py", line 75, in main
    d = import_module(splitext(sys.argv[1])[0])
  File "/usr/lib64/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
ImportError: Import by filename is not supported.
```

"d" was good name for a dictionary, but here we have a drive module, so maybe we should cal lit "mod"?

It would be nice to add a test for this code, probably extracting the driver loading code to a separate function (e.g. load_driver_module()), and testing that loading a driver file with relative and absolute path return the expected object.

rose/client/main.py
```diff
@@ -1,4 +1,6 @@
 import sys
+from os.path import splitext
+from importlib import import_module
```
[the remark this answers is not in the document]
yoavtzelnick: importlib is python builtin as well as os.path.
Could you please explain what do you mean?

rose/client/main.py
yoavtzelnick: I updated the pull request.
unfortunately python doesn't support import a file without destroying the PYTHONPATH.
the only way to make this work is to make examples a python package with __init__.py which I don't think make sense.
Is running a driver file from a different folder a requirement?

rose/client/main.py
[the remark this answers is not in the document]
nirs: We like to import only modules, not names from modules. The imports should be:
```
import os
import importlib
```

rose/client/main.py
nirs: Yes, I usually run drivers from another repo like this:

    ./rose-client ../rose-dev/score.py

Modifying sys.path or requiring that the driver modules are inside a package is ok, but this tool is for young kids and I don't want to make it more complicated then needed. Providing a relative or absolute path to a file should work.

We use imp to implement this:
https://docs.python.org/2/library/imp.html

yoavtzelnick: Great suggestion!
Updated the pull request.
Also made sure the parameter is a python file (source or compiled).

rose/client/main.py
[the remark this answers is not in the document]
yoavtzelnick: Done

rose/client/main.py
```diff
@@ -72,8 +72,18 @@ def main():
...
+    module_name = os.path.basename(module_path)
+
+    if file_suffix not in (".py", ".pyc"):
+        print "driver file must be either .py or .pyc"
+        sys.exit(3)
+
+    if file_suffix == ".py":
+        driver_mod = imp.load_source(module_name, file_name)
+    else:
+        driver_mod = imp.load_compiled(module_name, file_name)
```
nirs: This mostly works (does not support .pyo files), but why bother with the different file formats, when we can use the higher level apis like imp.find_module() and imp.load_module()? In this case imp module does the right thing for us.

See https://docs.python.org/2/library/imp.html#examples

Also, the code for loading the driver module should be extracted to a separate function to make it easy to test.

yoavtzelnick: 👍 Thanks for the productive feedback.
Now using imp.find_module and imp.load_module.
I extracted to load_driver_module, I'm not sure though if the expected behavior of this function is to return None or raise an exception.

yoavtzelnick:
All tests are passing.
Seems like this commit 058334bd8ad2ea0769accce07a4f4f1aeee09719 breaks build (Add CI dependency for PEP8 although not all code passes)

rose/client/main.py
```diff
@@ -68,16 +70,40 @@ def send_message(self, msg):
         self.client.sendLine(str(msg))
 
 
+def load_driver_module(file_path):
+    """
+    Title Try to load the driver module from file
```
nirs: Title is not needed, sorry if my example text was confusing.

Also, "Try to" is not needed. If may be good we had an interesting handling of errors, trying to load the driver in various ways, but we don't.

rose/client/main.py
```diff
@@ -68,16 +70,40 @@ def send_message(self, msg):
...
+
 def main():
     logging.basicConfig(level=logging.INFO)
     if len(sys.argv) < 2:
         log.info('usage: rose-client drive-module')
         sys.exit(2)
+    try:
+        driver_mod = load_driver_module(sys.argv[1])
+    except ImportError:
+        print "Couldn't load driver file"
```
nirs: We want to include the path of the file and the text of the error. This will make it easier to debug the issue when a user is complaining about a failure - see the example in my comment in a previous version of this PR.

rose/client/main.py
```diff
@@ -1,4 +1,6 @@
+import imp
 import logging
+import os.path
```
nirs: We can replace this with "import os".

nirs:
@yoavtzelnick, sorry about breaking the build, it should be fixed once we merge #238.

rose/client/main.py
yoavtzelnick: os.path is the precise module. We are not using os

Issue #190, opened 2017-10-19
Convert prints to logging


nirs:
Replace the non-portable print calls with logging.

moshe742:
hacktoberfest tlv shotgun

Abelarm:
I could work on this if noone already started

nirs:
@Abelarm @moshe742 started to work on the rose.server package - do you want to fix this in the rose.client and/or rose.common package?

Abelarm:
sure i'll start with client

nirs:
Thanks @Abelarm, waiting for your PR :+1:

Abelarm:
@nirs sorry for late but I can only code at night and weekends...
take a look and tell me if are there any change you need me to do 👍

nirs:
We stil have some prints in the client package and in the rose-admin tool.

Waiting for a contributor to do the last clean ups :-)

nirs:
@moshe742 do you want to complete this issue?

PR #223, opened 2017-10-21
fixed pep8/flake8 errors and enabled flake8 in travis-ci

eric-horvat:
Closes #218

PR #216, opened 2017-10-20
replace prints with logging in rose.server package

moshe742:
see #190

PR #225, opened 2017-10-21
replaced the prints with logs (rose/client - rose/common)

Abelarm:
Modified the client package, and the common one (common didn't have prints) 

see #190

### DIFF
--- a/rose/client/car.py
+++ b/rose/client/car.py
@@ -11,7 +11,6 @@ class Car(component.Component):
         self.id = id
         self.x = None
         self.y = None
-        self.texture = None
         self.name = None
 
     def update(self, info):
--- a/rose/client/car.py
+++ b/rose/client/car.py
@@ -1,9 +1,4 @@
-import logging
-import random
-import os
-from rose.common import config
 import component
-
 
 class Car(component.Component):
 
--- a/rose/client/car.py
+++ b/rose/client/car.py
@@ -4,7 +4,6 @@ import os
 from rose.common import config
 import component
 
-log = logging.getLogger('car')
 
 class Car(component.Component):
 
--- a/rose/client/car.py
+++ b/rose/client/car.py
@@ -1,7 +1,10 @@
+import logging
 import random
 import os
 from rose.common import config
 import component
+
+log = logging.getLogger('car')
 
 class Car(component.Component):
 
--- a/rose/client/component.py
+++ b/rose/client/component.py
@@ -1,7 +1,3 @@
-import logging
-
-log = logging.getLogger('Component')
-
 class Component(object):
     """ Component intarface """
 
--- a/rose/client/component.py
+++ b/rose/client/component.py
@@ -1,3 +1,7 @@
+import logging
+
+log = logging.getLogger('Component')
+
 class Component(object):
     """ Component intarface """
 
--- a/rose/client/game.py
+++ b/rose/client/game.py
@@ -1,3 +1,4 @@
+import logging
 import time
 from twisted.internet import reactor
 
@@ -8,6 +9,7 @@ import world
 import component
 
 author = 'gickowic'
+log = logging.getLogger('game')
 
 
 class Game(component.Component):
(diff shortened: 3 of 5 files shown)


rose/client/component.py
```diff
@@ -1,3 +1,7 @@
+import logging
+
+log = logging.getLogger('Component')
```
nirs: If we don't log anything here, we don't need logging yet.

rose/client/car.py
```diff
@@ -1,8 +1,11 @@
+import logging
 import random
 import os
 from rose.common import config
 import component
 
+log = logging.getLogger('car')
```
nirs: If we don't log anything here, we don't need logging yet.

rose/client/game.py
```diff
@@ -60,18 +62,18 @@ def car(self):
     # Handling client events
 
     def client_connected(self):
-        print 'client connected: joining as', self.name
+        log.info('client connected: joining as', self.name)
```
nirs: When converting prints to logging, we need format string and args - for example:

    log.info('client connected: %s', self.name)

rose/client/main.py
```diff
@@ -25,7 +27,7 @@ def lineReceived(self, line):
         elif msg.action == 'error':
             self.factory.error(msg.payload)
         else:
-            print 'Unexpected message:', msg.action, msg.payload
+            log.info('Unexpected message:', msg.action, msg.payload)
```
nirs: Same, need proper format string.

rose/client/main.py
```diff
@@ -67,7 +69,7 @@ def send_message(self, msg):
 
 def main():
```
nirs: We need to configure the logger here - see #216

rose/client/track.py
```diff
@@ -1,8 +1,10 @@
 import glob
+import logging
 import os
 from rose.common import config, obstacles
 import component
 
+log = logging.getLogger('main')
```
nirs: We don't need logging here yet.

rose/client/car.py
```diff
@@ -1,5 +1,10 @@
+import logging
```
nirs: logging is not needed here.

rose/client/car.py
```diff
@@ -1,5 +1,10 @@
...
+import random
+import os
+from rose.common import config
```
nirs: These imports were just removed on master and are not needed after #219 and #227.

rose/client/car.py
```diff
@@ -1,5 +1,10 @@
...
 import component
 
+
```
nirs: This is correct, but not related to import changes - other people are working on pep8 fixes, and such change will conflict with their changes, see #223.

rose/client/game.py
```diff
@@ -60,18 +62,19 @@ def car(self):
...
+        log.info('client disconnected: %s', reason.getErrorMessage())
 
     def client_failed(self, reason):
-        print 'client failed:', reason.getErrorMessage()
+        log.info('client failed: %s', reason.getErrorMessage())
 
     def client_error(self, error):
-        print 'client error:', error.get('message')
+        log.info('client error: %sp', error.get('message'))
+        log.info('client error: %sp', error.get('message'))
```
nirs: Two issues:
- log line duplicated
- trailing "p" at the end of the format string.

rose/client/game.py
```diff
@@ -60,18 +62,18 @@ def car(self):
...
-        print 'client disconnected:', reason.getErrorMessage()
+        log.info('client disconnected: %s', reason.getErrorMessage())
 
     def client_failed(self, reason):
-        print 'client failed:', reason.getErrorMessage()
+        log.info('client failed: %s', reason.getErrorMessage())
 
     def client_error(self, error):
-        print 'client error:', error.get('message')
+        log.info('client error: %sp', error.get('message'))
```
nirs: There is still trailing "p" in the format string

Abelarm: you're right just amending the commit

Abelarm: sorry 👎

PR #227, opened 2017-10-21
Remove last traces of pygame

nirs:
Some imports and instance variables are not needed now in the client
package, after pygame was removed in #219.

### DIFF
--- a/rose/client/car.py
+++ b/rose/client/car.py
@@ -1,6 +1,3 @@
-import random
-import os
-from rose.common import config
 import component
 
 class Car(component.Component):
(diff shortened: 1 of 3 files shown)


nirs:
@Ditti4 can you review these changes, completing #219?

rose/client/car.py
```diff
@@ -9,7 +6,6 @@ def __init__(self, id, x, y):
         self.id = id
         self.x = None
         self.y = None
```
Ditti4: One off-topic question arises here for me: why not

- do away with the `x` and `y` arguments in the `__init__` or
- assign the values now that we have them.

Obviously this is not related to your change, I only noticed it just now.

nirs: @Ditti4 right, the x, y arguments are useless and can be removed - why not send another PR?

Issue #232, opened 2017-10-21
Reversed sorting in server stats


nirs:
When a game completes, the server print the stats to the console:
```
Stats
1     Zig Zag  row:5  score:620
2       Score  row:2  score:707
```
The order of the drivers is wrong - should be sorted by score, highest score first.

PR #234, opened 2017-10-21
Fix of score printing

Abelarm:
according to #232

### DIFF
--- a/rose/server/game.py
+++ b/rose/server/game.py
@@ -94,7 +94,7 @@ class Game(object):
     def print_stats(self):
         print
         print 'Stats'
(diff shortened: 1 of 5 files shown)


rose/server/game.py
```diff
@@ -94,7 +94,7 @@ def drive_player(self, name, info):
     def print_stats(self):
         print
         print 'Stats'
-        for i, p in enumerate(sorted(self.players.values())):
+        for i, p in enumerate(sorted(self.players.values(), reverse=True)):
```
nirs: @Abelarm thanks, looks correct.

Abelarm: 👍

Issue #231, opened 2017-10-21
Mute/unmute the sound in the web interface


nirs:
Some people (e.g. me) do not like the soundtrack. It will be useful to mute the sound easily from the web UI without muting the entire machine.

The UI should probably be at the top of the window near the other elements, something like a laptop mute button maybe.

The system should keep the mute state; closing and opening the browser will keep the sound muted.

I think html5 local storage api is the best way to implement this, since we don't have any concept of authentication in this application.
See https://html.spec.whatwg.org/multipage/webstorage.html#storage-2

Issue #230, opened 2017-10-21
Add sound in the web UI


nirs:
In the old pygame code we played a sound from ```rose/res/soundtrack/Nyan_Cat.ogg``` when starting a client.

We like to restore this functionality by playing this sound in a loop when opening the web interface.

PR #236, opened 2017-10-22
Add web soundtrack

devmattrick:
Adds a soundtrack to the webpage upon startup (also allows the ability to mute/ unmute and stores it in local storage), resolving both #230 and #231.

Please let me know if you'd like any changes made!

### DIFF
--- a/rose/web/game.js
+++ b/rose/web/game.js
@@ -22,6 +22,9 @@ var ROSE = (function() {
         }.bind(this));
 
         var soundLoader = new SoundLoader(function() {});
+        soundLoader.load("res/soundtrack/Nyan_Cat.ogg", function(sound) {
+            sound.play();
+        });
 
         this.context = $("#game").get(0).getContext("2d");
         this.dashboard = new Dashboard(imageLoader);
@@ -29,7 +32,6 @@ var ROSE = (function() {
         this.obstacles = new Obstacles(imageLoader);
         this.cars = new Cars(imageLoader);
         this.finish_line = new FinishLine(imageLoader);
-        this.sound_controller = new SoundController(soundLoader);
     }
 
     App.prototype.onmessage = function(m) {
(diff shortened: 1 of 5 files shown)


nirs:
@devmattrick thanks for your contribution! Looks great so far, will review it later.

rose/web/game.js
```diff
@@ -17,16 +17,21 @@ var ROSE = (function() {
         this.controller = new Controller();
         this.rate = new Rate([0.5, 1.0, 2.0, 5.0, 10.0]);
 
-        var loader = new ImageLoader(function() {
+        var imageLoader = new ImageLoader(function() {
```
nirs: I would like to avoid mixedCase names in our code, they are more error prone and harder to read and write. Can rename this to image_loader?

rose/web/game.js
```diff
@@ -405,6 +403,38 @@ var ROSE = (function() {
...
+            true: "res/icons/unmute.svg"
+        };
+
+        var self = this;
+
+        $("#toggle_mute img").attr('src', icons[this.muted]);
+
+        loader.load("res/soundtrack/Nyan_Cat.ogg", function(sound) {
+            sound.muted = self.muted;
+            sound.loop = true;
```
nirs: Don't we need this in the previous commit? The sound should always loop, regardless of having a way to mute it.

rose/web/game.js
```diff
@@ -405,6 +403,38 @@ var ROSE = (function() {
...
+            sound.play();
+            self.sound = sound;
+        });
+
+        $("#toggle_mute").click(function(event) {
+            event.preventDefault();
+
+            self.muted = !self.muted;
+            $(this).children('img').attr('src', icons[self.muted]);
```
nirs: I think we should use:

    $("#toggle_sound img").attr(...)

rose/web/index.html
```diff
@@ -14,6 +14,8 @@
         <button id="start" disabled>Start</button>
         <button id="stop"  disabled>Stop</button>
 
+        <button id="toggle_mute"><img src="res/icons/mute.svg"></button>
```
nirs: This possibly starts with the wrong icon, I wonder if we should have a third icon for the loading step, or make this button disabled until we read the state from local storage.

devmattrick:
Thanks for the suggestions, I'll work on it as soon as j get a chance!

rose/web/game.js
```diff
@@ -400,6 +402,28 @@ var ROSE = (function() {
...
+        var self = this;
+
+        loader.load("res/soundtrack/Nyan_Cat.ogg", function (sound) {
+            sound.loop = true;
+            self.sounds[0] = sound;
+            self.toggle_sound(0);
+        });
+    }
+
+    SoundController.prototype.toggle_sound = function(sound_id) {
```
nirs: I'm not sure how you are going to use this in the next version, since when clicking sound button, we have to:
- play/pause sound
- replace the icon

It seems that the entire sound controller belongs to the next PR, this PR is happy with the plain loader, playing the file when it was loaded, as in the first commit.

nirs:
@devmattrick thanks for this fine work! I hope you can find time to send the next PR for #231.

Issue #135, opened 2015-01-27
Add good samples for training 


emesika:
We need to create new samples based on teh score.py and zigzag.py that can be given as a pyc files for students for trainign purpose.
Those can be used by us as well to test that game and find bugs ....

PR #240, opened 2017-10-23
make random-driver names more random

ReneVolution:
While preparing for a different Todo on the project, I came across that random_driver.py is not really random and can't be run twice. So, I changed the approach of naming the drivers differently.

Hope this works for you. The good about this is for people just wanting to quickly run this project and see what it does, it's easy to just use the random driver twice.

What you think?

### DIFF
--- a/examples/random-driver.py
+++ b/examples/random-driver.py
@@ -7,8 +7,23 @@ very good driver but the implementation is very elegant.
 import random
 from rose.common import obstacles, actions
 
+attribute = [
+    "Furious",
+    "Big",
+    "Rolling"
+]
+
+names = [
+    "Mike",
+    "Sally",
+    "Jackson",
+    "Sam",
+    "Jane",
+    "Bruce"
+]
+
 server_address = "localhost"
-driver_name = "Random Driver"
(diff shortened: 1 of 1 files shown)


nirs:
Hi @ReneVolution, this is very fun idea, I like it!

But I don't think it is a good idea for the example drivers. The consumers of this project are young kids doing their first steps in Python, and this make the driver module more complicated to understand and does not help them to write a driver module.

Solving the problem of running a driver module twice is creating another driver file, or copying a driver file from another student.

We plan to include compiled examples driver modules that can be used by students or developers, see #224 and #135.

See https://github.com/nirs/rtp-dev for example drivers for development. Please do not publish them, we do not want to spoil the game for new students.

Maybe you will be interested to work on this:
https://github.com/treyhunner/names

ReneVolution:
Hey @nirs,

thanks a lot for your feedback. You're right, the fun it brings may make it too complex for the kids to get. I'll have a look at the resources you provided and follow up later.

nirs:
@ReneVolution thanks for your PR!

PR #243, opened 2017-10-25
Add examples compiled drivers for training

### LINKED ISSUE #135: Add good samples for training

nirs:
- score.pyc - driver checking the obstacles multiple rows a head and
  picking the action that will give the best score.
- zigzag.pyc - driver checking single row a head, and generally like to
  turn right and left.

Fixes #135.


nirs:
@emesika, @rollandf, can you review?

PR #249, opened 2017-10-26
Pipenv integration

### LINKED ISSUE #228: Complete pipenv integration, remove requirements duplication
We added pipenv configuration, but we still pip in travis, and we have duplicate requirements files.

Changes needed:
- Use pipenv in travis instead pip
- Remove requirements.txt

(Generating requirements.txt move to #257)

rodxavier:
- [x] Update .travis.yml

- [x] Remove requirements.txt

- [x] Generate requirements.txt in python setup.py sdist

Fixes #228

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,19 @@
+import subprocess
+import sys
 from distutils.core import setup
-from subprocess import call
+
+
+DATA_FILES = []
 
 
 def generate_requirements():
     with open('requirements.txt', 'w') as f:
-        call(['pipenv', 'lock', '--requirements'], stdout=f)
+        subprocess.check_call(['pipenv', 'lock', '--requirements'], stdout=f)
     return 'requirements.txt'
+
+
+if 'sdist' in sys.argv:
+    DATA_FILES.append(('requirements.txt', generate_requirements()))
 
 
 setup(name='rose-project',
@@ -18,4 +26,4 @@ setup(name='rose-project',
       author_email='ybronhei@redhat.com',
       url="https://github.com/emesika/RaananaTiraProject",
       scripts=["rose-client", "rose-server", "rose-admin"],
-      data_files=[('requirements.txt', [generate_requirements()]), ])
+      data_files=DATA_FILES)
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,12 @@
 from distutils.core import setup
+from subprocess import call
+
+
+def generate_requirements():
+    with open('requirements.txt', 'w') as f:
+        call(['pipenv', 'lock', '--requirements'], stdout=f)
+    return 'requirements.txt'
+
 
 setup(name='rose-project',
       version='0.1',
@@ -9,4 +17,5 @@ setup(name='rose-project',
       author='Yaniv Bronhaim',
       author_email='ybronhei@redhat.com',
       url="https://github.com/emesika/RaananaTiraProject",
-      scripts=["rose-client", "rose-server", "rose-admin"],)
+      scripts=["rose-client", "rose-server", "rose-admin"],
+      data_files=[('requirements.txt', [generate_requirements()]), ])
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 htmlcov/
 
 Pipfile.lock
+requirements.txt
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -1,4 +1,12 @@
 from distutils.core import setup
+from subprocess import call
+
+
+def generate_requirements():
+    with open('requirements.txt', 'w') as f:
+        call(['pipenv', 'lock', '--requirements'], stdout=f)
```
nirs: call() will fail silently on errors, we would like to fail loudly in this case.

setup.py
```diff
@@ -9,4 +17,5 @@
       author='Yaniv Bronhaim',
       author_email='ybronhei@redhat.com',
       url="https://github.com/emesika/RaananaTiraProject",
-      scripts=["rose-client", "rose-server", "rose-admin"],)
+      scripts=["rose-client", "rose-server", "rose-admin"],
+      data_files=[('requirements.txt', [generate_requirements()]), ])
```
nirs: This will generate the file for every setup.py run - can we do this only for sdist?

nirs:
@rodxavier one more comment, with this change, we should not mention ```pip install -r requirements.txt``` in README.md.

rodxavier:
@nirs Made the following changes

- remove requirements.txt in readme
- Replace `call()` with `check_call()` to not fail silently
- Only generate requirements when doing `sdist`. Honestly, I'm not sure if there's a more elegant way to do this.

setup.py
```diff
@@ -1,13 +1,21 @@
...
 
 def generate_requirements():
     with open('requirements.txt', 'w') as f:
-        call(['pipenv', 'lock', '--requirements'], stdout=f)
+        subprocess.check_call(['pipenv', 'lock', '--requirements'], stdout=f)
     return 'requirements.txt'
 
 
+if 'sdist' in sys.argv:
+    DATA_FILES.append(('requirements.txt', generate_requirements()))
```
nirs: Returning "requirements.txt" from generate_requirements() is little hacky. generate_requirements should not have a return value, we call it for the side effect.

If we need to add the tuple ("requirements.txt", "requirements.txt") specifying it as is it simpler and more clear.

I think the correct way to do this is to create our own sdist class, and create the generated files before we run the real sdist.

Basically this is something like:
```
from distutils.command import sdist


class rose_sdist(sdist.sdist):

    def run(self):
        self.genrate_files()
        sdist.sdist.run(self)

    def generate_files(self):
        # generate the requirements file here...

detup(
    cmdclass = {"sdist": "rose_sdist"},
    ...
)
```
See how this is done for overriding the install_data class in https://wiki.python.org/moin/Distutils/Tutorial.

We can also generate the file as part of the "build" command - doesn't "sdist" depend on "build"?

setup.py
```diff
@@ -18,4 +26,4 @@ def generate_requirements():
       author_email='ybronhei@redhat.com',
       url="https://github.com/emesika/RaananaTiraProject",
       scripts=["rose-client", "rose-server", "rose-admin"],
-      data_files=[('requirements.txt', [generate_requirements()]), ])
+      data_files=DATA_FILES)
```
nirs: Can we specify data_files always, and generate the file only when running sdist?

nirs:
@rodxavier since generating requirements.txt require much more work, do you want to send another PR for this, so we can merge now the rest of the changes?

rodxavier:
@nirs that sounds good. will open another PR for the requirements stuff.

nirs:
@rodxavier thanks, I'll merge manually most of the commits here. Please rebase the new PR after I merge.

nirs:
@rodxavier thanks!

I merged 3 commit from this PR in:
- 7145300ed9c039f8dfaa9618072e4319841b7bd5
- 4b924d3c4c6f4a2fcd337a756d847bbdb0ab762a
- 642b7e367a0dad71e9e78961ab427d07e6cadd01

I also amended the commit messages to match the project style, and changed the pipenv install command to use only long options, makes the code more readable.

Strangely, the build fail now - see:
https://travis-ci.org/RedHat-Israel/ROSE/builds/293907888

Can you take a look? I don't have experience with running pipenv in travis.

nirs:
@rodxavier, I sent #258 to fix the build issue, so this PR is done. Thanks!

PR #252, opened 2017-10-27
Replace print with logging in rose-admin and rose.client

### LINKED ISSUE #190: Convert prints to logging

moshe742:
Fixes #190.

### DIFF
--- a/rose-admin
+++ b/rose-admin
@@ -16,6 +16,7 @@ class Client(object):
         self.server = None
 
     def run(self):
+        logging.basicConfig(level=logging.INFO)
         if len(self.args) < 2:
             self.usage('server-address required')
 
@@ -35,7 +36,7 @@ class Client(object):
         try:
             method()
         except Exception as e:
-            log.warning(e)
+            log.error("Error running: %s, %s", verb, e)
             sys.exit(1)
 
     def do_start(self):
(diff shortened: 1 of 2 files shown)


rose/client/main.py
```diff
@@ -101,7 +101,7 @@ def main():
     try:
         driver_mod = load_driver_module(sys.argv[1])
     except ImportError as e:
-        print "error loading driver module %r: %s" % (sys.argv[1], e)
+        log.warning("error loading driver module %r: %s" % (sys.argv[1], e))
```
nirs: We want to pass the arguments to the logger instead of formatting them before calling the logger. Also, this is an error, should use log.error().

rose-admin
```diff
@@ -31,7 +35,7 @@ class Client(object):
         try:
             method()
         except Exception as e:
-            print e
+            log.warning(e)
```
nirs: log.error("Error running %s: %s", verb, e)

rose-admin
```diff
@@ -2,9 +2,13 @@
 
 import sys
 import os
+import logging
 import xmlrpclib
 from rose.common import config
 
+log = logging.getLogger('admin')
+
```
nirs: We need to configure the logger before we use it. Can be done in Client.run().

PR #258, opened 2017-10-27
Avoid Pipfile.lock in travis

### LINKED ISSUE #228: Complete pipenv integration, remove requirements duplication

nirs:
Travis fails now when trying to lock the packages. Avoid the unneeded
locking step and use the Pipfile instead.

Fixes #228.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
 
 install:
     - pip install pipenv
-    - pipenv install --dev --system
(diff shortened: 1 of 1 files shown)


nirs:
@rodxavier can you review this fix?

PR #260, opened 2017-10-28
Added requiremetns.txt

prashasy:


nirs:
@prashasy thanks for your work, but we just removed requirement.txt. We are using pipenv to manage packages.

PR #261, opened 2017-10-29
Add comment to generated requirements.txt

### LINKED ISSUE #257: Generate requirements.txt for building a tarball
We like to publish the project on pypy (see #189), without requiring pipenv, so people that do not use pipenv can install the project using pip.

Tasks:
1. Generate requirements.txt from Pipenv when running python setup.py sdist
2. Make sure the file is mentioned in the MANIFEST
3. Make sure git does not complain about the unknown file in "git status"

nirs:
Fixes #257.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,8 @@ class rose_sdist(sdist.sdist):
 
     def generate_files(self):
         with open('requirements.txt', 'w') as f:
+            f.write("# Generated automatically from Pipfile - do not edit!\n")
+            f.flush()
             subprocess.check_call(['pipenv', 'lock', '--requirements'],
                                   stdout=f)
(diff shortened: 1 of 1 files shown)


rodxavier:
Just need to change Pipefile to Pipfile in the comment.

nirs:
@rodxavier thanks, will fix on next version.

setup.py
```diff
@@ -10,6 +10,8 @@ def run(self):
 
     def generate_files(self):
         with open('requirements.txt', 'w') as f:
+            f.write("# Generated automatically from Pipefile - do not edit!\n")
```
nirs: Pipefile -> Pipfile.

PR #269, opened 2017-10-31
Improve logging format

### LINKED ISSUE #268: Improve logging format
Current logging format is not very useful:

Client:
```
$ ./rose-client -s 10.35.0.197 examples/score.pyc 
INFO:main:Please open http://10.35.0.197:8880 to watch the game.
```
Server:
```
$ ./rose-server 
INFO:main:starting server
INFO:game:add player: u'Score', lane: 1, car: 2
```

I think in both cases we like to have something like:

```
2017-10-31 19:00:00,000 INFO  [logger] message ... (func:lineno)
```

obernardovieira:
Fixes #268.

### DIFF
--- a/rose/server/game.py
+++ b/rose/server/game.py
@@ -9,8 +9,7 @@ from . import track
 from . import player
 from . import score
 
-FORMAT = '%(asctime)-15s %(levelname)s [%(name)-8s] %(message)s (%(funcName)s:%(lineno)d)'
-logging.basicConfig(format=FORMAT)
+logging.basicConfig(format=config.logger_format)
 log = logging.getLogger('game')
 
 
--- a/rose/server/game.py
+++ b/rose/server/game.py
@@ -9,6 +9,8 @@ from . import track
 from . import player
 from . import score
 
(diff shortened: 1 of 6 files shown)


rose/server/game.py
```diff
@@ -9,6 +9,8 @@
 from . import player
 from . import score
 
+FORMAT = '%(asctime)-15s %(levelname)s [%(name)-8s] %(message)s (%(funcName)s:%(lineno)d)'
```
nirs: Some issues in the format:
- We don't need to set the timestamp width or change the default alignment
- The levelname width differ, we like to align it so the next fields are aligned
  the widest level name is "WARNING", so we need -%7s
- We don't need to align the logger name
- In the issue I described (func:lineno), but I meant module:lineno, sorry for the confusion
- We use the same format in rose-server, rose-client, and rose-admin - so we want to keep one copy of the format, best in rose/common/config.py

rose/server/game.py
```diff
@@ -9,6 +9,8 @@
...
+logging.basicConfig(format=FORMAT)
```
nirs: We don't need to add logging.basicConfig - this is done once in rose/server/main.py, rose/client/main.py, and rose-admin.

nirs:
Thanks @obernardovieira!

Issue #263, opened 2017-10-30
Avoid useless game without players


rollandf:
If all the drivers disconnect, the game still runs for no reason.
The game should stop.

nirs:
I think we can solve it differently - when one player disconnects, the other player wins, and the game stops.

If we go in this way, starting a game should be possible only when 2 players are connected.

emesika:
+1

rollandf:
I have seen a lot of students running their driver only while testing their code.
I don't think that adding a limitation of two client is useful.

nirs:
@rollandf, good point about single user usage for training.

So I think the changes needed are:
- frontend: Enable the "Start" button only when at least one player is connected
- frontend: Render "No player connected" message if no player is connected.
- backend: fail start() if no player is connected
- backend: When a player disconnects, stop the game if it was the last player

nirs:
Fixed by #273.

PR #273, opened 2017-11-14
Stop game if no players connected

rollandf:
Fixes issue #263 

-  frontend: Enable the "Start" button only when at least one player is connected
- frontend: Render "No player connected" message if no player is connected.
- backend: fail start() if no player is connected
- backend: When a player disconnects, stop the game if it was the last player

### DIFF
--- a/rose/web/game.js
+++ b/rose/web/game.js
@@ -44,7 +44,7 @@ var ROSE = (function() {
         var state = msg.payload;
 
         // Update
-        this.controller.update(state.started);
+        this.controller.update(state);
         this.rate.update(state.rate);
         this.dashboard.update(state);
         this.track.update(state);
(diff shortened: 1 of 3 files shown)


rose/web/game.js
```diff
@@ -44,7 +44,7 @@ var ROSE = (function() {
         var state = msg.payload;
 
         // Update
-        this.controller.update(state.started);
+        this.controller.update(state.started, state.players.length);
```
nirs: Better pass entire state, so we can have a uniform update(state) - will allow later to replace this code with a loop calling update on all components.

rose/web/game.js
```diff
@@ -44,7 +44,7 @@ var ROSE = (function() {
...
         this.rate.update(state.rate);
```
nirs: (This should be fixed later to accept state as well)

rose/web/index.html
```diff
@@ -13,6 +13,7 @@
 
         <button id="start" disabled>Start</button>
         <button id="stop"  disabled>Stop</button>
+        <label id="info"></label>
```
nirs: This works, but I thought abut something more fancy - instead of rending the track/obstances/cars - render big gray "NO PLAYER CONNECTED". This require much more work, so lets do this later.

If you like to play with this idea, I think it can be done like this:

- in all components, do not render anything if state.players.length == 0
- in a new NoPlayer component, render the the text when state.players.length == 0 and render nothing otherwise.

Another option is to have multiple divs, one for the game canvas, one for "no player connected" message, and show only one div at a time using hidden="true" and hiden="none". The the text can be plain html <h2>No player connected</h2>

rose/web/game.js
rollandf: Done

PR #292, opened 2018-02-25
Python content: add Exceptions

rollandf:


docs/course_materials/exceptions.html
```diff
@@ -0,0 +1,309 @@
...
+          </code></pre>
+        </section>
+        <section>
+          <h3><b>Accessing the Exception</b></h3>
+           <ul>
+            <li>Use "as" to access the exception object.</li>
+          </ul>
+<pre><code data-trim data-noescape class="hljs">
+try:
+    raise RuntimeError("o hai")
```
sleviim: what?

docs/course_materials/exceptions.html
```diff
@@ -0,0 +1,309 @@
...
+    print('Outer')
+          </code></pre>
+            <ul>
+            <li>The (default) root exception handler terminates the Python process.</li>
+          </ul>
+        </section>
+        <section>
+          <h3><b>Propagating Exceptions</b></h3>
+          <ul>
+            <li>Propagation can be forced by using raise without arguments</li>
```
sleviim: maybe 'raise' , or bolding

docs/course_materials/exceptions.html
rollandf: Changed to "oops"

docs/course_materials/exceptions.html
rollandf: changed to 'raise'

PR #300, opened 2018-07-14
Convert examples/README to markdown for nicer rendering on GitHub

cben:
before: https://github.com/RedHat-Israel/ROSE/tree/master/examples
after: https://github.com/cben/ROSE/tree/patch-1/examples


nirs:
It is nicer, thanks.

PR #302, opened 2018-07-15
README: Execute commands under pipenv

cben:
Without it, they don't use the virtualenv and don't find dependencies:
```
$ ./rose-server
  File "./rose-server", line 3, in <module>
    from rose.server import main
  File "/home/bpaskinc/ROSE/rose/server/main.py", line 4, in <module>
    from twisted.internet import reactor
ImportError: No module named twisted.internet
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ The race car must recognize the race track, the obstacles, and the bonus areas;
 then calculate the best path where the pitfalls are avoided and all the bonus points are collected.
 The cars move autonomously on the screen within the race track game with no interference
 from the students. No joystick or mouse shall be used.
- 
+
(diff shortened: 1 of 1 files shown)


cben:
@nirs I reworded for `pipenv shell` as you suggested, PTAL.

However given the use case of 2 or 3 terminals, you don't really save any typing from `pipenv shell` (in fact you type 3 chars more ;-), and you risk getting weird errors if you forgot `pipenv shell`.
IMHO prepending `pipenv run ` to each command is more bullet-proof.

cben:
@nirs can get this merged before the students use these instructions tomorrow?

nirs:
Thanks Beni!

PR #309, opened 2018-10-02
ROSE Server in a Docker Container

### LINKED ISSUE #173: ROSE Server container
Create a container for the server and publish it on docker hub.

With the container, instructor and/or students can run a game using docker without installing anything but docker.

ecliptik:
Fixes #173

Uses official CentOS Python 2.7 image.

Verified building and running locally and can access ROSE Server on http://locahlost:8880

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+# Use official CentOS Python 2.7 image
+# https://developer.fedoraproject.org/tools/docker/docker-images.html
+FROM centos/python-27-centos7 AS base
+
+# Project maintainer
+LABEL maintainer="frolland@redhat.com"
+
+# Required env vars
+ENV LD_LIBRARY_PATH="/opt/rh/python27/root/usr/lib64/" \
+    ENABLE_PIPENV=true
+
+# Install pipenv
+RUN pip install --upgrade pipenv
+
+# Copy application to default WORKDIR
+COPY . ./
+
+# Install dependencies from pipfile
+RUN pipenv install
+
+# Server port
+EXPOSE 8880
+
+# Server command
+ENTRYPOINT [ "pipenv" ]
+CMD [ "run", "python", "rose-server"]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# Project maintainer
+LABEL maintainer="frolland@redhat.com"
+
+# Required env vars
+ENV LD_LIBRARY_PATH="/opt/rh/python27/root/usr/lib64/" \
+    ENABLE_PIPENV=true
+
+# Install pipenv
+RUN pip install --upgrade pipenv
```
nirs: When we used a container, we don't need pipenv.

Dockerfile
```diff
@@ -0,0 +1,26 @@
+# Use official CentOS Python 2.7 image
+# https://developer.fedoraproject.org/tools/docker/docker-images.html
+FROM centos/python-27-centos7 AS base
+
+# Project maintainer
+LABEL maintainer="frolland@redhat.com"
+
+# Required env vars
+ENV LD_LIBRARY_PATH="/opt/rh/python27/root/usr/lib64/" \
+    ENABLE_PIPENV=true
```
nirs: Why do we need LD_LIBRARY_PATH?

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# Install pipenv
+RUN pip install --upgrade pipenv
+
+# Copy application to default WORKDIR
+COPY . ./
+
+# Install dependencies from pipfile
+RUN pipenv install
```
nirs: We can install the requirements manually, or using a requirements.txt file generated
by pipenv.

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# Server port
+EXPOSE 8880
+
+# Server command
+ENTRYPOINT [ "pipenv" ]
+CMD [ "run", "python", "rose-server"]
```
nirs: And without pipeenv, we can just run rose-server here.

Issue #295, opened 2018-03-05
UI : Do not play sound by default in the browser


rollandf:
The music is played automatically when the page is loaded.
It is very distracting when teaching about the game.

It should not be playing by default.
I suggest that until we have a mute button in the UI, the music should not be played.

rollandf:
Closed by #330

PR #310, opened 2018-10-03
Disable sound in the browser

rollandf:
Disabling the music in the browser until we are able to mute it.
See issue #295


rollandf:
ping

PR #313, opened 2018-12-10
Rename meet_linux.md to linux_intro.md

sleviim:


rollandf:
Thanks!
I fixed in #312

PR #323, opened 2019-11-21
Python 3.7 Porting

aabawer:
Patchset for porting the ROSE project into python 3.7, targeted for Fedora 30+.
rose-server/client/admin mods were tested over fc30 machine using the random example players.


rollandf:
:+1:  LGTM

@nirs  can you look at it?

PR #327, opened 2019-12-24
Using input on python2

sleviim:
This patch adds an example of using strings values from input, and
casting the input's value to an int.

### DIFF
--- a/docs/course_materials/variables_data_types.html
+++ b/docs/course_materials/variables_data_types.html
@@ -494,11 +494,17 @@ a = (3 + 2)*5 + 6/3 #  5*5 + 2 =27
               <li>input()</li>
               <li>input(prompt) : prints prompt before reading input</li>
             </ul>
-            <li>The result is of String type</li>
           </ul>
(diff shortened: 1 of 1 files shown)


docs/course_materials/variables_data_types.html
```diff
@@ -494,11 +494,17 @@ <h3><b>Interactive Input</b></h3>
               <li>input()</li>
               <li>input(prompt) : prints prompt before reading input</li>
             </ul>
-            <li>The result is of String type</li>
           </ul>
           <pre><code data-trim data-noescape class="hljs">
-name = input('Enter your name: ')
-age = int(input('Enter your age: '))
+>>> name = input('Enter your name: ')
```
nirs: In python 2 you should use ```raw_input()```,
see https://docs.python.org/2/library/functions.html#input

sleviim: It seems that in python2, raw_input casts automatically to a string, while input needs to enter a value with ''.
Which one is better for your opinion?

PR #330, opened 2020-07-21
Add music control to UI

EgorLu:
Hi, these are the changed I've made after looking at [issue:295](https://github.com/RedHat-Israel/ROSE/issues/295) and [issue:231](https://github.com/RedHat-Israel/ROSE/issues/231).
- Music is not playing automatically when the page loads.
- Can be toggled via UI or JS API.

### DIFF
--- a/rose/web/game.css
+++ b/rose/web/game.css
@@ -35,7 +35,7 @@ body {
     background-color: #222;
 }
 
-#start, #stop {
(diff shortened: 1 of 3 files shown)


EgorLu:
I've tried to stick to the JS style of the document.
Please tell me what you think.

rollandf:
Thanks!

Issue #332, opened 2020-07-29
Recompile the example drivers


EgorLu:
Currently the example driver files don't run.
My assumption is that with the transition to Python 3, the driver files need to be recompiled.

If someone still has their source code, it's a minute's work. Otherwise they could be reverse-engineered / decompiled back into source code, but that might be an overkill.

rollandf:
Hi,
Can you tell me which ones, I will look if I have the sources.
Thanks

EgorLu:
Hi Fred,
It's [score.pyc](https://github.com/RedHat-Israel/ROSE/blob/master/examples/score.pyc) and [zigzag.pyc](https://github.com/RedHat-Israel/ROSE/blob/master/examples/zigzag.pyc)

PR #334, opened 2020-08-02
Convert and recompile example drivers to Python3

EgorLu:
* Recompiled to Python 3.7.8
* Checked that both drivers work


rollandf:
:100:  Thanks!

EgorLu:
Resolves #332

PR #339, opened 2020-08-23
WIP: Add support of configurable ports

rollandf:
Still needed:
- Update documentation
- Handle `rose-admin` (https://github.com/RedHat-Israel/ROSE/blob/master/rose-admin#L31)


rollandf:
Closing this as we are not going to implement this for now.

PR #341, opened 2020-08-23
Add support for configurable ports in rose-admin

shiramax:


rollandf:
Closing this as we are not going to implement this for now.

PR #351, opened 2020-10-08
Updated the scoring method

AnnaDinaburgVulikh:
Fixed the punishment score so it reduces 10 points
Removed extra forward step for pickup
Added an if statement for the case of unmagical obstacles

Checked the scoring between the drivers after the fix, seems more fair.
There is a notable difference between a driver that tries to follow the rules and a random one.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -69,6 +69,8 @@ with the lane owner.
     |       |       |       |X|       |       |      |
     |       |       |       |X|       |       |      |
 
+Every turn the driver moves one step forward and gains 10 points
+if there was no obstacle.  
 Here are the possible obstacles and the action you should take when hitting
 them:
 
@@ -76,35 +78,35 @@ them:
                         action.
                         Score : 0
 
-* `obstacles.PENGUIN`   return `actions.PICKUP` and move 1 step forward. You will
-                        not move backward if you do not pick up the penguin.
+* `obstacles.PENGUIN`   return `actions.PICKUP` and move forward. You will
+                        move forward even if you do not pick up the penguin.
                         Score : 10 for `PICKUP` otherwise 0
 
 * `obstacles.WATER`     return `actions.BRAKE` if you want to continue forward.
                         return `actions.RIGHT` or `actions.LEFT` if you like to
                         bypass the obstacle. Any other action will cause you to
-                        move backward.
+                        stay in place.
                         Score: 4 for `BRAKE`, 0 for `LEFT`/`RIGHT` otherwise -10
 
 * `obstacles.CRACK`     return `actions.JUMP` if you want to continue forward
                         return `actions.RIGHT` or `actions.LEFT` if you like to
                         bypass the obstacle. Any other action will cause you to
-                        move backward.
+                        stay in place.
                         Score: 5 for `JUMP`, 0 for `LEFT`/`RIGHT` otherwise -10
 
 * `obstacles.TRASH`     you must return `actions.RIGHT` or `actions.LEFT` to bypass
                         the obstacle. Any other action will cause you to
-                        move backward.
+                        stay in place.
                         Score: 0 for `LEFT`/`RIGHT` otherwise -10
 
 * `obstacles.BIKE`      you must return `actions.RIGHT` or `actions.LEFT` to bypass
                         the obstacle. Any other action will cause you to
-                        move backward.
+                        stay in place.
                         Score: 0 for `LEFT`/`RIGHT` otherwise -10
 
 * `obstacles.BARRIER`   you must return `actions.RIGHT` or `actions.LEFT` to bypass
                         the obstacle. Any other action will cause you to
-                        move backward.
+                        stay in place.
                         Score: 0 for `LEFT`/`RIGHT` otherwise -10
 
 NOTE : If you are driving out of your lane and have a collision with the lane
@@ -112,4 +114,4 @@ NOTE : If you are driving out of your lane and have a collision with the lane
 
 
 Please check and test drive the examples drivers modules in this directory.
-Then go ahead and write the fastest driver!
+Then go ahead and write the best driver!
--- a/rose/server/score.py
+++ b/rose/server/score.py
@@ -40,27 +40,26 @@ def process(players, track):
             if player.action != actions.JUMP:
                 track.clear(player.x, player.y)
                 player.y += 1
-                player.score += config.score_move_backward
+                player.score += config.score_move_backward * 2
             else:
                 player.score += config.score_jump
-
         elif obstacle in (obstacles.TRASH,
                           obstacles.BIKE,
                           obstacles.BARRIER):
-            track.clear(player.x, player.y)
-            player.y += 1
-            player.score += config.score_move_backward
+            if player.action != actions.LEFT and player.action != actions.RIGHT:
+                track.clear(player.x, player.y)
+                player.y += 1
+                player.score += config.score_move_backward * 2
         elif obstacle == obstacles.WATER:
             if player.action != actions.BRAKE:
                 track.clear(player.x, player.y)
                 player.y += 1
-                player.score += config.score_move_backward
+                player.score += config.score_move_backward * 2
             else:
                 player.score += config.score_brake
         elif obstacle == obstacles.PENGUIN:
             if player.action == actions.PICKUP:
                 track.clear(player.x, player.y)
-                player.y -= 1
                 player.score += config.score_move_forward
 
         # Here we can end the game when player gets out of


sleviim (changes requested):
@rollandf is there something we shall take into considuration?

Anna, this one requires changing the README for the scoring as well

AnnaDinaburgVulikh:
Do you mean adding the fact that moving forward grants points?
Will add, just have to mention that it was the scoring method till now.
I only fixed the penalty score and the extra move for pickup.

sleviim:
I meant that on https://github.com/RedHat-Israel/ROSE/blob/master/examples/README.md you have a detailed explanation about each obstacle and its action.
It says, for example, "obstacles.PENGUIN return actions.PICKUP and move 1 step forward. You will not move backward if you do not pick up the penguin. Score: 10 for PICKUP otherwise 0"

If there's no longer one step forward it should be documented

AnnaDinaburgVulikh:
I'm not sure this should be changed. In the game now, each 'step' is forward unless you tell it otherwise since the track is moving on. Before the fix above, 'pickup' made the driver jump 2 steps forward and score 20 points while doing it, the same as just moving forward 2 steps without doing anything special. The fix gives the driver 20 points for one step and 'pickup'.
Moreover, if you don't tackle obstacles as 'bike' and so on you actually stay in place instead of moving forward with the track.
If the goal is the **fastest** driver, than the 'pickup' should give 2 steps and score 30 points correspondingly, 20 for 2 steps forward and 10 for the pickup.
Please let me know what scoring method you prefer and the goal of the game - fastest driver or best score.

sleviim:
The score method is by the greatest score (that way we can take the out of 3 wins).
And about the README, I'm a bit confused...
IIUC, there's no penalty for moving backward but there's a -10 point reduces?
If so the documentation should be changed then.

AnnaDinaburgVulikh:
You are correct, you stay in place even thou the track moves forward.
The real meaning of the current penalty, is that you can't get the points back if you decided not to handle the obstacles.
Updated the README.

sleviim (approved):
LGTM,
@rollandf, can you please have a look to see we didn't miss something?

shiramax (approved):
thanks @AnnaDinaburgVulikh  it makes sense!

shiramax:
@rollandf It would be great if you could take a look at this PR :)

nirs:
I don't understand the problem, and how the changes fix it.
Whats wrong with the current rules of the game?

examples/README.md
```diff
@@ -69,47 +69,49 @@ with the lane owner.
     |       |       |       |X|       |       |      |
     |       |       |       |X|       |       |      |
 
+Every turn the driver moves one step forward and gains 10 points
+if there was no obstacle.
```
nirs, line 73: This looks like a good change, making the behavior more clear

examples/README.md
```diff
@@ -69,47 +69,49 @@ with the lane owner.
...
 them:
 
 * `obstacles.NONE`      there is no obstacle at this position. You can return any
                         action.
                         Score : 0
 
-* `obstacles.PENGUIN`   return `actions.PICKUP` and move 1 step forward. You will
-                        not move backward if you do not pick up the penguin.
+* `obstacles.PENGUIN`   return `actions.PICKUP` and move forward. You will
+                        move forward even if you do not pick up the penguin.
```
nirs, line 82: Why did you remove "1 step forward"? Getting extra points and movement is intended. This is a game about Linux, so picking up penguins should be special.

Having obstacles that have higher priority makes the game more interesting and challenging. For example if you have 2 obstacles you can handle in future move, player that give priority to the penguin should be more successful.

examples/README.md
```diff
@@ -69,47 +69,49 @@ with the lane owner.
...
                         Score : 10 for `PICKUP` otherwise 0
 
 * `obstacles.WATER`     return `actions.BRAKE` if you want to continue forward.
                         return `actions.RIGHT` or `actions.LEFT` if you like to
                         bypass the obstacle. Any other action will cause you to
-                        move backward.
+                        stay in place.
```
nirs, line 88: Why "stay in place" is better than move "backwards"?

examples/README.md
```diff
@@ -69,47 +69,49 @@ with the lane owner.
...
                         Score: 0 for `LEFT`/`RIGHT` otherwise -10
 
 NOTE : If you are driving out of your lane and have a collision with the lane
        owner, you will punished by -10 points and be moved to other free place.
 
 
 Please check and test drive the examples drivers modules in this directory.
-Then go ahead and write the fastest driver!
+Then go ahead and write the best driver!
```
nirs, line 117: This is a race, usually the fastest driver is the best.

rose/server/score.py
```diff
@@ -40,27 +40,26 @@ def process(players, track):
             if player.action != actions.JUMP:
                 track.clear(player.x, player.y)
                 player.y += 1
-                player.score += config.score_move_backward
+                player.score += config.score_move_backward * 2
```
nirs, line 43: Why is it better to give more score?
Why not change the config so config.score_move_backward is larger?

I don't think this is a good change, it makes configuration harder since
we need to consider this magic hidden factor.

It is better to change the configuration instead of the code.

rose/server/score.py
```diff
@@ -40,27 +40,26 @@ def process(players, track):
...
             else:
                 player.score += config.score_jump
-
```
nirs, line 46: Unrelated change

rose/server/score.py
```diff
@@ -40,27 +40,26 @@ def process(players, track):
...
         elif obstacle in (obstacles.TRASH,
                           obstacles.BIKE,
                           obstacles.BARRIER):
-            track.clear(player.x, player.y)
-            player.y += 1
-            player.score += config.score_move_backward
+            if player.action != actions.LEFT and player.action != actions.RIGHT:
```
nirs, line 49: LEFT and RIGHT are already handled. Why do we need to consider 
them again?

rose/server/score.py
```diff
@@ -40,27 +40,26 @@ def process(players, track):
...
                 track.clear(player.x, player.y)
                 player.y += 1
-                player.score += config.score_move_backward
+                player.score += config.score_move_backward * 2
             else:
                 player.score += config.score_brake
         elif obstacle == obstacles.PENGUIN:
             if player.action == actions.PICKUP:
                 track.clear(player.x, player.y)
-                player.y -= 1
```
nirs, line 63: This makes the penguin less special.

examples/README.md
AnnaDinaburgVulikh: This change actually makes it. Before picking up the penguin would have given you 20 points and 2 steps. The same as just moving forward.

Moreover, there were situations where the extra move puts you on an obstacle and you 'lose' points since you don't get the chance to handle it.

rose/server/score.py
AnnaDinaburgVulikh: It was intended to cancel out the score for moving forward, they are the same. Will fix to removing the forward score.

rose/server/score.py
AnnaDinaburgVulikh: The condition is on line 49, before the handling that starts on line 50.

AnnaDinaburgVulikh: @nirs Before the change, for bike, trash and barrier, there was no check for the action chosen by the driver. Each time those obstacles were met the driver was losing points and staying in place (since the track moves forward, and the driver moves down, he actually stays in place).
I added the condition and moved the actions under it.

examples/README.md
AnnaDinaburgVulikh: Another option is keeping the extra step, clearing the field after the penguin, adding 20 points for forward (2 steps) + adding 10 points for penguin pickup +? adding points for removed obstacle (can be another penguin).

In any case, keeping the current score doesn't give the advantage for picking up the penguin.

shiramax:
Hi @nirs, thanks for the review!
The reason I requested the change from @AnnaDinaburgVulikh, a volunteer that contributes time and code to the ROSE project, is that in the past years we didn't see a big change in the scores of a student that is algorithm is and great and a student that is algorithm is not that great, sometimes even the not-great one won,
We believe that the reasons are that you get 10 points for just moving forward, and don't get 'punishments' for not dealing with obstacles,

Please let us know if you think this solution is helpful, or if you think of other solutions that might work.

shiramax:
@nirs @sleviim @rollandf can we merge it?

shiramax:
Thanks @AnnaDinaburgVulikh for your contribution. and thanks @sleviim and @nirs for the review. I'm Merging this one.

PR #353, opened 2020-10-10
Added an automation for checking exercise 1 in linux intro

AnnaDinaburgVulikh:
Please let me know if this is the direction you wanted.
If so, I will make the automations for the other exercises.

I ran it in the terminal using `pipenv shell` and 'python check_exe_1.py`


shiramax (changes requested):
Hi @AnnaDinaburgVulikh looks good!!
a few comments :
1. please change the file name to be check_class_exercise_1
2. please add a comment for how to run this automation check (maybe you can add an option --help to the script that provides information to how to run the script, I guess each check could be run differently )
3. what happens if a student did well the exercise and just keep continue to write more then 20 commands?

AnnaDinaburgVulikh:
Hi @shiramax , I'm working on a more general solution.
An executable file, that will get the path and run the check exercise file as a module, like the driver.
It works but I get a Deprecation Warning for the `imp` module.
Should I switch to the `importlib` (introduced for python 3.1)?
How far should the backward compatibility go, which python version?  There are function for python 3.4 for example.

shiramax:
@AnnaDinaburgVulikh we are using python 3.7 in the course. so you can use importlib or other functions :)

AnnaDinaburgVulikh:
Hi @shiramax,
1. Updated the name of the file as you asked.
2. Made an executable to run the exercise check files, similar to the rose-driver caller.
3. Updated the log to include 100 commands, so the check would be fast, but allows a room for some writing.
4. Not sure which README should include the instructions for the rose-check executable and whether the location of the file should be in the exercise folder. Could you advise?

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
```diff
@@ -0,0 +1,51 @@
...
+from os import path
+from pathlib import Path
+
+home = str(Path.home())
+exercise_commands = ['mkdir test', 'mkdir tmp', 'tree']
+
+def get_student_commands():
+    '''
+    Checks stutent history file for key commands.
+    If the history is longer than 20 commands, checks only 20 last commands.
```
shiramax, line 12: 100

shiramax, line 12: 100

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
```diff
@@ -0,0 +1,51 @@
+import argparse
+import os.path
+from os import path
+from pathlib import Path
+
+home = str(Path.home())
+exercise_commands = ['mkdir test', 'mkdir tmp', 'tree']
+
```
shiramax, line 8: what if mkdir will be run with some additional arguments,
mdkir -p test/tmp  

also, what if the command runs correctly but wasn't created? or created in the wrong place? 
I think that in this case, it's enough to check if the directories exist

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
```diff
@@ -0,0 +1,51 @@
...
+    Checks stutent history file for key commands.
+    If the history is longer than 20 commands, checks only 20 last commands.
+    '''
+    my_commands = []
+    hist_path = home + '/.bash_history'
+    with open(hist_path, 'r') as rf:
+        for line in rf:
+            my_commands.append(line[:-1])
+    if len(my_commands) > 100:
+        my_commands = my_commands[-100:]
```
shiramax, line 20: why not 1000?

docs/course_materials/exercises/rose-check.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+
```
shiramax, line 2: Looks great!

shiramax (changes requested):

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
```diff
@@ -0,0 +1,51 @@
...
+    '''
+    test_result = ''
+    student_commands = get_student_commands()
+    if not used_right_commands(student_commands):
+        test_result += '''It seems you missed some commands.
+    Make sure you have used 'mkdir' for creating the directories
+    and 'tree' for viewing the directory tree.\n'''
+    if not created_directories():
+        test_result += '''It seems that not all directories were created in the right place.
+    Make sure you create a 'test' directory in 'home'
```
sleviim, line 47: a -> the

docs/course_materials/exercises/rose-check.py
```diff
@@ -0,0 +1,52 @@
...
+            
+            
+'''
+Calling the corresponding check_exercise file.
+'''
+logging.basicConfig(level=logging.INFO)
+parser = argparse.ArgumentParser(description="ROSE Exercise")
+parser.add_argument("exercise_file",
+                    help="The path to the check_exercise file."
+                         " For example: '01_Linux/check_class_exercise_1.py'")
```
sleviim, line 46: extra space before the For

sleviim (changes requested):

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
AnnaDinaburgVulikh: I wanted to check for mkdir to make sure the terminal was used. 
If you think it's not needed I will remove it.

AnnaDinaburgVulikh:
@shiramax @sleviim made the changes you requested.

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
```diff
@@ -0,0 +1,50 @@
...
+    test_result = ''
+    student_commands = get_student_commands()
+    if not used_right_commands(student_commands):
+        test_result += ("It seems you missed some commands. "
+                    "Make sure you have used 'mkdir' for creating the directories "
+                    "and 'tree' for viewing the directory tree.\n")
+    if not created_directories():
+        test_result += ("It seems that not all directories were created in the right place. "
+                    "Make sure you create the 'test' directory in 'home' "
+                    "and a 'tst' directory in 'test'.")
```
shiramax, line 47: same here

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
```diff
@@ -0,0 +1,50 @@
...
+
+def main():
+    '''
+    Running the tests and prompting the student according to his work.
+    '''
+    test_result = ''
+    student_commands = get_student_commands()
+    if not used_right_commands(student_commands):
+        test_result += ("It seems you missed some commands. "
+                    "Make sure you have used 'mkdir' for creating the directories "
```
shiramax, line 42: I think we can generate this message to write exactly which command wasn't run properly.

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
```diff
@@ -0,0 +1,50 @@
+import argparse
+import os.path
+from os import path
+from pathlib import Path
+
+home = str(Path.home())
+exercise_commands = ['mkdir', 'tree']
+
+def get_student_commands():
```
shiramax, line 9: thanks @AnnaDinaburgVulikh 
one more thing, if we want to use these functions in other exercises, I think it's better to extract all these functions to a helper file: 
* get_student_command
* used_right_command
*created_directories

shiramax (changes requested):

AnnaDinaburgVulikh:
@shiramax made the changes

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
```diff
@@ -0,0 +1,28 @@
+import os.path
```
shiramax, line 1: I believe you don't need this here anymore

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
```diff
@@ -0,0 +1,28 @@
...
+from pathlib import Path
```
shiramax, line 2: same here

docs/course_materials/exercises/check_helper.py
```diff
@@ -0,0 +1,35 @@
...
+from pathlib import Path
+
+home = str(Path.home())
+
+def get_student_commands():
+    '''
+    Checks stutent history file for key commands.
+    If the history is longer than 1000 commands, checks only 1000 last commands.
+    '''
+    my_commands = []
```
shiramax, line 11: I tried to run the script, I don't have this file, maybe we can use the history command instead? 
and I was thinking, maybe instead of getting the last 1000 command, maybe we can check commands from a range of dates? like for example for the past 10 days? what do you think?

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
```diff
@@ -0,0 +1,28 @@
...
+
+def main():
+    '''
+    Running the tests and prompting the student according to his work.
+    '''
+    test_result = ''
+    student_commands = check_helper.get_student_commands()
+    unused_commands = check_helper.find_unused_commands(student_commands, exercise_commands)
+    if len(unused_commands) > 0:
+        test_result += "It seems you missed some commands. Make sure you have used the following commands:\n"
```
shiramax, line 17: here I meant it could be great to present exactly which command was missed, or which exactly directory wasn't created.

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
```diff
@@ -0,0 +1,28 @@
...
+home = str(Path.home())
+exercise_commands = {'mkdir': 'creating directories', 'tree': 'viewing the directory tree'}
+exercise_paths = [home + '/test/tmp']
+
+def main():
+    '''
+    Running the tests and prompting the student according to his work.
+    '''
+    test_result = ''
+    student_commands = check_helper.get_student_commands()
```
shiramax, line 14: I think we can have another function, test_exercise or something similar, and all of what is in main could be there, and this function could be in the helper file as well, I believe you will use this function for other exercises as well.

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
```diff
@@ -0,0 +1,28 @@
+import os.path
+from pathlib import Path
+import check_helper
+
+home = str(Path.home())
```
shiramax, line 5: I think it could a global variable, maybe create a config file and add this variable so we won't have this twice in two places.

shiramax (changes requested):

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
AnnaDinaburgVulikh: Made the 'home' a global variable in check_helper, this file would be used for all the exercises.

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
AnnaDinaburgVulikh: It's doing that for the commands.
For the path, I didn't break it down to pieces. It's possible to add more paths to the check.

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
AnnaDinaburgVulikh: done

docs/course_materials/exercises/check_helper.py
AnnaDinaburgVulikh: Made the adjustments, it should run properly now. Please let me now if there are problems.
(Had a problem at first to since I'm using zsh)

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
AnnaDinaburgVulikh: Done

AnnaDinaburgVulikh:
Hi @shiramax, added checks for hw1 & 2 in Linux and made the changes we talked about in the helper file.
Added a check for deleted diles/directories.

I think there is a problem in hw2, it says to move `welcome.txt`, but it wasn't created.
There are also no instructions for using `head`, `tail` and `diff`.

docs/course_materials/exercises/01_Linux/check_homework_1.py
```diff
@@ -0,0 +1,31 @@
+from pathlib import Path
+import check_helper
+
+exercise_commands = {
+    'pwd': 'printing current directory name',
+    'mkdir': 'creating directories',
+    'cd': 'changing between directories',
+    'ls': 'listing directory contents',
+    'tree': 'viewing the directory tree',
```
shiramax, line 9: I  think it's better to have a full dictionary in the helper of entire commands and their description to not repeat ourselves, and in here you can just have an array with the names of the commands that are relevant to the specific exercise.

docs/course_materials/exercises/01_Linux/check_homework_2.py
```diff
@@ -0,0 +1,31 @@
...
+    'man': 'opening manual',
+    'mkdir folder1 folder2 folder3': 'creating multiple folders',
+    'welcome.txt ~/hw1/other': 'moving welcome file (partial command)', 
+    'welcome.txt ~/hw1/docs/backup.txt': 'copying welcome file to backup.txt (partial command)'
+    } 
+
+exercise_paths = [check_helper.HOME + '/hw1/other/folder1',
+                check_helper.HOME + '/hw1/other/folder2',
+                check_helper.HOME + '/hw1/other/folder3',
+                check_helper.HOME + '/hw1/other/welcome.txt',
```
shiramax, line 20: Maybe inside test_exercise you can add the HOME directory for each directory in the array? so we won't have this repetition?

docs/course_materials/exercises/check_helper.py
```diff
@@ -0,0 +1,91 @@
+import os
+from pathlib import Path
+import datetime as dt
+
```
shiramax, line 4: this exercise will run from the root but it should run on /home/sutudent. any chance you can add an option to change the home directory?

docs/course_materials/exercises/check_helper.py
```diff
@@ -0,0 +1,91 @@
...
+HOME = str(Path.home())
+'''
+Setting colors for the output:
+green for positive feedback
+red negative feedback
+yellow for commands and paths, without new line
+'''
+def prGreen(feedback): print("\033[92m {}\033[00m" .format(feedback)) 
+def prRed(feedback): print("\033[91m {}\033[00m" .format(feedback))
+def prYellow(feedback): print("\033[93m {}\033[00m" .format(feedback), end='')
```
shiramax, line 14: why prYellow and not print_yellow?

shiramax (changes requested):

docs/course_materials/exercises/check_helper.py
AnnaDinaburgVulikh: Added as an option for the run command

docs/course_materials/exercises/01_Linux/check_homework_2.py
```diff
@@ -1,24 +1,15 @@
...
+exercise_commands = ['cp', 'mv', 'head', 'tail', 'diff', 'clear', 'man',
+                    'mkdir folder1 folder2 folder3', 'welcome.txt ~/hw1/other', 'welcome.txt ~/hw1/docs/backup.txt'
+                    ]
 
-exercise_paths = [check_helper.HOME + '/hw1/other/folder1',
-                check_helper.HOME + '/hw1/other/folder2',
-                check_helper.HOME + '/hw1/other/folder3',
-                check_helper.HOME + '/hw1/other/welcome.txt',
-                check_helper.HOME + '/hw1/docs/backup.txt'
+exercise_paths = ['/hw1/other/folder1',
```
shiramax, line 8: I would also remove from here the first /  and add it in the function . 
because these are a relative path.

docs/course_materials/exercises/01_Linux/check_homework_1.py
```diff
@@ -1,27 +1,15 @@
...
-    'docs': 'a folder in hw1 (partial command)'
-    } 
+exercise_commands = ['pwd', 'mkdir', 'cd', 'ls', 'tree', 'rm', 'touch', 'cat', 'vim', 'website.txt', 'docs'] 
 
-exercise_paths = [check_helper.HOME + '/hw1',
-                check_helper.HOME + '/hw1/web',
-                check_helper.HOME + '/hw1/other',
+exercise_paths = ['/hw1',
+                '/hw1/web',
+                '/hw1/other',
```
shiramax, line 8: I would also remove from here the first / and add it in the function .
because these are a relative path.

docs/course_materials/exercises/check_helper.py
```diff
@@ -50,12 +75,12 @@ def created_all_paths(exercise_paths):
     num_of_created_paths = 0
     print("Checking created directories or files...")
     for cur_path in exercise_paths:
-        prYellow("'" + cur_path + "'")
+        print_yellow("'" + HOME + cur_path + "'")
         print(" ... ", end='')
-        if not os.path.exists(cur_path):
-            prRed("Missing")
+        if not os.path.exists(HOME + cur_path):
```
shiramax, line 80: created_all_paths and deleted_all_paths content is pretty much the same, can you extract the content to a function?

shiramax (changes requested):

shiramax:
@AnnaDinaburgVulikh I also change the homework a bit based on your comments. 
Also, can you add a readme that explains how to add a new test for a Linux exercise? 
thanks!

AnnaDinaburgVulikh:
> @AnnaDinaburgVulikh I also change the homework2 a bit based on your comments:
> #366
> 
> Also, can you add a readme that explains how to add a new test for a Linux exercise?
> thanks!

Hi @shiramax , made all the changes including adding the README.

AnnaDinaburgVulikh:
Hi @shiramax , made the changes we talked about.

AnnaDinaburgVulikh:
@shiramax done, including new class_exercises.

docs/course_materials/exercises/rose_check.py
```diff
@@ -0,0 +1,82 @@
...
+    if module_spec is None:
+        print('Module: {} not found'.format(module_name))
+        return None
+    else:
+        try:
+            module = importlib.util.module_from_spec(module_spec)
+            module_spec.loader.exec_module(module)
+            return module
+        except FileNotFoundError as e:
+            print("Error loading the file %r: %s" % (file_path, e.strerror))
```
EgorLu, line 36: Two things about string formatting:
1. Since we're using Python 3, let's use f-strings :) You're using C-style formatting here and `format` method above.
2. We should be consistent with quotes. You've used single quotes previously, so let's stick to that.

docs/course_materials/exercises/linux_tester.py
```diff
@@ -0,0 +1,110 @@
...
+# set from rose_check.py from the check_exercise file.
+HOME = '' 
+COMMANDS = []
+PATHS = []
+DELETED_PATHS = []
+
+
+def get_student_commands():
+    """
+    Checks stutent history file for key commands in the past 10 days.
```
EgorLu, line 15: Typo
```suggestion
    Checks student history file for key commands in the past 10 days.
```

AnnaDinaburgVulikh:
Hi @EgorLu, made the changes you suggested.

AnnaDinaburgVulikh:
> i think it can be moved to the line before

@sleviim made the fix.

shiramax (commented):
Hi @AnnaDinaburgVulikh ,
When running the script, I'm getting this error: 
  File "rose_check.py", line 75, in <module>
    exercise_mod = load_file(args.exercise_file)
  File "rose_check.py", line 26, in load_file
    module_spec = importlib.util.spec_from_file_location(module_name,
AttributeError: module 'importlib' has no attribute 'util'

I investigate a bit: 

`
>>> import importlib
>>> importlib.util
  File "<stdin>", line 1, in <module>
AttributeError: module 'importlib' has no attribute 'util'
>>> from importlib import util
>>> util

importlib.util
<module 'importlib.util' from '/usr/lib64/python3.6/importlib/util.py'>

`
 I don't know why one works and the other doesn't.

docs/course_materials/exercises/01_Linux/check_class_exercise_1.py
```diff
@@ -0,0 +1,11 @@
+from pathlib import Path
```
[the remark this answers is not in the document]
AnnaDinaburgVulikh: Was removed

AnnaDinaburgVulikh:
Hi @shiramax , added setup instruction to readme (needs python 3.7) and moved everything to a test_exercises folder.

shiramax (approved):
Great work! thanks anna!

PR #375, opened 2020-10-28
Add UTF-8 encoding to HTML

EgorLu:
Previously encoding was unset, which resulted in an error in the console and potentially wrong encoding with various browsers.

### DIFF
--- a/rose/web/index.html
+++ b/rose/web/index.html
@@ -1,7 +1,8 @@
-<!doctype html>
+<!DOCTYPE html>
 <html>
 
   <head>
+    <meta charset="UTF-8">
     <title>ROSE Game</title>
     <link rel="stylesheet" type="text/css" href="game.css">
   </head>


sleviim (commented):
Hi Egor,
Do you have an ecample for one of the failures you've mentioned?

EgorLu:
Well, generally it's advised to explicitly set the encoding in an HTML document.

Firefox, for example, yells:
```
The character encoding of the HTML document was not declared. The document will render with garbled text in some browser configurations if the document contains characters from outside the US-ASCII range. The character encoding of the page must be declared in the document or in the transfer protocol. localhost:8880
```

EgorLu:
https://imgur.com/uYtY7xs
This is the screenshot of the error

sleviim:
Thanks for the detailed error and solution :)

PR #398, opened 2020-11-09
change format to f-string format

mgold1234:
This PR can be merged only after the end of ROSE current cycle.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -25,4 +25,4 @@ twisted = "*"
 
 [requires]
 
-python_version = "3.7"
+python_version = "3.6.8"
--- a/rose-admin
+++ b/rose-admin
@@ -28,8 +28,9 @@ class Client(object):
             self.usage('server-address required')
 
         server_address = self.args[1]
-        url = 'http://%s:%d/rpc2' % (server_address, config.web_port)
+        url = f'http://f{server_address}:{config.web_port}/rpc2'
         self.server = xmlrpclib.Server(url)
+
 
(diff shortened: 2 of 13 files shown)


shiramax:
@mgold1234 can you please rebase?

Pipfile
```diff
@@ -25,4 +25,4 @@ twisted = "*"
 
 [requires]
 
-python_version = "3.7"
+python_version = "3.6.8"
```
rollandf, line 28: Please revert

rollandf:
@mgold1234 tests are failing

rose-admin
```diff
@@ -39,11 +40,11 @@ class Client(object):
         try:
             method = getattr(self, method_name)
         except AttributeError:
-            self.usage('Unknown verb: %r' % verb)
+            self.usage(f'Unknown verb: {verb}')
         try:
             method()
         except Exception as e:
-            log.error("Error running: %s, %s", verb, e)
+            log.error(f"Error running: {verb}, {e}")
```
nirs, line 47: This is bad change. Logging message should not be formatted. Formatting the messages means we always pay for formatting the logs even when the logs are not enabled (e.g. log.debug() when log level is WARNING).

log.error() is likely always loged, but it does not make sense to handle differently different log messages, so logging calls always pass the arguments to the logger and the logger format the messages before they are written.

Please remove all the changes in logging calls.

nirs (changes requested):

PR #406, opened 2020-11-19
score.py: avoid failing on flake8 test

sleviim:
The line's length is currently 80>79 chars.
This patch tries to fix it, so the Travis test won't fail.


AnnaDinaburgVulikh:
@sleviim looks great, I'll adopt this to future code.

shiramax:
@sleviim can you plase rebase?

PR #407, opened 2020-11-19
Fixed the test according to the new score system.

AnnaDinaburgVulikh:
@sleviim , @shiramax Fixed the score_test.py according to the changes in the scoring method.
All test passed.

### DIFF
--- a/rose/server/score_test.py
+++ b/rose/server/score_test.py
@@ -43,12 +43,12 @@ class SinglePlayerTest(object):
     def assert_move_back(self):
         assert self.player.x == self.x
         assert self.player.y == self.y + 1
-        assert self.player.score == self.score
+        assert self.player.score == self.score - config.score_move_forward
 
     def assert_move_back_no_punish(self):
         assert self.player.x == self.x
         assert self.player.y == self.y + 1
-        assert self.player.score == self.score
+        assert self.player.score == self.score - config.score_move_forward
 
     def assert_keep_obstacle(self):
         assert self.track.get(self.x, self.y) == self.obstacle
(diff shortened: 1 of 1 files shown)


rose/server/score_test.py
```diff
@@ -43,12 +43,12 @@ def assert_move_left(self):
     def assert_move_back(self):
         assert self.player.x == self.x
         assert self.player.y == self.y + 1
-        assert self.player.score == self.score
+        assert self.player.score == self.score - config.score_move_forward
```
sleviim, line 46: I've seen in the test, can someone get a <0 score?
like -10?
I think we should stay at 0 in that case.
maybe use something like `max(self.score - config.score_move_forward, 0)` for avoiding a negative score

AnnaDinaburgVulikh: Each move you get 10 points for moving forward, then your score is modified according to the action you chose.
The first 5 steps are clear of obstacles so you can say that the first obstacle is reached with 60 points.
In those conditions, I suspect that a score lower than 0 won't happen.

AnnaDinaburgVulikh: Even a no action driver finishes the race with a score higher than 250.

sleviim: I was wondering about those due to the failures I got in my patch:
https://travis-ci.org/github/RedHat-Israel/ROSE/builds/744630508
(After solving the flake8 issue and start running the tests)

shiramax:
thanks @AnnaDinaburgVulikh, it looks great!, @sleviim thanks for the review

Issue #427, opened 2021-04-28
Docer for ROSE


AnnaDinaburgVulikh:
Hello,
While trying to resolve #173 , encountered a problem with installing the DEV dependencies on Docker.
Here is the conflict:
![image](https://user-images.githubusercontent.com/66639090/116424845-8659a400-a84a-11eb-90ae-648b248b34fe.png )

sleviim:
@nirs @rollandf 
Do you have any inssights about that one?

AnnaDinaburgVulikh:
@nirs @rollandf 
Since the requirements.txt is generated from the pipfile, tried specifing the needed version in the pipfile, it didn't work.
Writing a designated requirements.txt with the needed version of `idna` worked.
The simple solution, as I see it, is generating a specific requirements file for DEV and setting the `idna` to the correct version. Should I set a specific version for other dependencies as `pytest` or just for `idna`? 
Or maybe am I missing another solution?

sleviim:
@nirs can you please share your thoughts on this one?

nirs:
We don't need pipenv to create a container. A container can replace pipenv completely.

PR #428, opened 2021-04-28
Dockerfile for ROSE

AnnaDinaburgVulikh:
Hi @sleviim @sfishbai , 
This pr is trying to follow the notes for #309 and raised the issue #427 .

One way of solving the issue is creating a requirements_dev.txt with the needed dependencies.
The question is, would you prefer to lock all the module versions or only `idna`(the conflicting one)?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,42 @@
-# Use official CentOS Python 2.7 image
-# https://developer.fedoraproject.org/tools/docker/docker-images.html
-FROM centos/python-27-centos7 AS base
+# Generate a container that generates requirements.txt
+ARG PY_VERSION=3.7
+FROM python:${PY_VERSION} as source
 
-# Project maintainer
-LABEL maintainer="frolland@redhat.com"
+ARG DEV
 
-# Required env vars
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,26 @@
+# Use official CentOS Python 2.7 image
```
sleviim: @rollandf , shoulds we use python 3.6 for it?

rollandf: Python 2.7 is not supported, so yes.

AnnaDinaburgVulikh: Hi, i'm not sure where you see this line.
In the files I commited the dicker uses python 3.7

sleviim: Maybe after rebasing it will be removed.
You can see another reference at line 3.

README.md
```diff
@@ -94,18 +94,32 @@ For running the same track for all drivers (instead or random) start the server
...
+
 ## Running a driver
 
 In a new window, open your virtual environment:
 
     pipenv shell
 
-Create your driver file: 
-    
+Create your driver file:
```
sleviim, line 117: redundent space attheend of the line

sleviim (changes requested):

sleviim (approved):

PR #433, opened 2021-11-09
Add recursion example

benukhanov:


benukhanov:
@sleviim 😄

docs/course_materials/functions.html
```diff
@@ -167,26 +167,25 @@ <h3><b>Example - is even? </b></h3>
...
-is_prime(2) # True
-is_prime(6) # False
-is_prime(7) # True
-is_prime(39) # False
-is_prime(31) # True
+my_print("ROSE", 5) # ROSE
+my_print("ROSE", 4) # ROSE
+my_print("ROSE", 3) # ROSE
+my_print("ROSE", 2) # ROSE
+my_print("ROSE", 1) # ROSE
```
sleviim, line 188: According to your example, the output for `my_print("ROSE", 5)` is 
ROSE
ROSE
ROSE
ROSE
ROSE

so lines 185-188 are redundant.

It also feels more like a for loop that prints the output X times, was that intended? 
or maybe use another example with the traditional factorials?

can add it to the extra, as i think it has more value

sleviim (changes requested):

docs/course_materials/functions.html
benukhanov: Yes, I want to show recursion in general ("an action that repeats itself until some condition is met").

The traditional factorial example requires knowledge of mathematics (I'm assuming the audience is not learning discrete mathematics). I don't want to clutter up the programming with mathematics, the example shouldn't take more than 2 minutes to explain.

https://en.wikipedia.org/wiki/Factorial

docs/course_materials/functions.html
```diff
@@ -167,26 +167,25 @@ <h3><b>Example - is even? </b></h3>
...
-    if x < 2:
-        return False
-    for n in range(2, x - 1):
-        if x % n == 0:
-            return False
-    return True
+def my_print(message, repeat_times):
+    if repeat_times == 0:
+        return
+
```
sleviim, line 178: Also, for clarity, can you please remove the empty line here? (178)

docs/course_materials/functions.html
```diff
@@ -167,26 +167,25 @@ <h3><b>Example - is even? </b></h3>
...
-def is_prime(x):
-    if x < 2:
-        return False
-    for n in range(2, x - 1):
-        if x % n == 0:
-            return False
-    return True
+def my_print(message, repeat_times):
+    if repeat_times == 0:
+        return
```
sleviim, line 177: note: the students might not be familiar with an empty 'return'.
need to be sure they are familiar with 'end term' as in while loops

sleviim (changes requested):
a very small one (review and a note)

docs/course_materials/functions.html
benukhanov: Done

docs/course_materials/functions.html
benukhanov: Done

sleviim (approved):
Thanks!

PR #441, opened 2022-07-07
Add ci workflow

nirs:
Started from python application template.


nirs:
This was an example, Josh should send a working PR.

PR #446, opened 2022-07-12
Ci

didib:


.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install flake8 pytest
+        pip install flake8 pytest pytest-cov pytest-timeout
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
nirs, line 30: It should work if we install as described in the readme:

    pip install pipenv
    pipenv install --dev

.github/workflows/ci.yml
```diff
@@ -0,0 +1,39 @@
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 pytest
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```
nirs, line 36: I would just run:

    flake8 --show-source --statistics .

Pipfile
```diff
@@ -26,4 +26,4 @@ idna = ">=2.5,<3"
 
 [requires]
 
-python_version = "3.7"
+python_version = "3.10"
```
nirs, line 29: This is not related to CI, should be fix regardless. Without this running pipenv install will install python 3.7 and "pipenv run" will run python 3.7 instead of current version.

nirs: I found that this does not work with the compiled drivers. We need to recompile them with python 3.10 if we want to upgrade python version, or remove them from the source.

nirs: I removed this change. We can upgrade python version later.

didib:
> I tried to use pipeenv but it is horribly slow for CI so I added simple requirments.txt files.
> 
> We can squash and merge to have working CI.
> 
> @didib what do you think?

OK for me.

For Nitzanim project last week, I did this on my own fork, and also guided my mentees do use their own fork.

Now that this week is over and we most likely have some time til the next use of current project, perhaps we can clean up everything including CI to use pipenv, but that's not mandatory - we can also do this later after merging and enabling CI.

nirs:
I think we should get rid of pipenv and use containers for CI, maybe also for development.
We have only 2 requirements (twisted, autobhan) and installing them using pip or distro packages is easy.

PR #448, opened 2022-07-12
Update README

yaacov:
While working on Nitazanim project we found some information we would like to update in the README, ( and related files ). 

  - [x] add `.vscode` to the gitignore list, in case a user want to use VSCode text editor
  - [x] use any python3, not just 3.7
  - [x] use podman instead of docker

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 
 # Editors
 .idea
+.vscode
 
 # Coverage
 .coverage
@@ -13,6 +14,5 @@ requirements.txt
 
 # Classroom setup
 classroom/credentials.json
-.vscode/launch.json
 classroom/*.csv
 classroom/token.pickle
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,13 @@
 init: Pipfile
-	pip install --user pipenv
+	python -m pip install pipenv --user
 	pipenv install
 
+dev-init: Pipfile
+	python -m pip install pipenv --user
+	pipenv install --dev
+
 test: pytest.ini
-	pytest
+	pipenv run pytest
 
 admin: rose-admin
 	pipenv run ./rose-admin
@@ -13,3 +17,6 @@ server: rose-server
 
 client: rose-client
 	pipenv run ./rose-client
+
+container-image:
+	podman build --build-arg DEV=True -t rose_dev .
--- a/README.md
+++ b/README.md
@@ -56,12 +56,7 @@ In order to make sure we have pipenv installed:
 If you don't have it installed, the best way is to install it only for
 your user:
 
-    pip install --user pipenv
-    
-Or with pip3:
-
-    pip3 install --user pipenv
-
+    python -m pip install --user pipenv
 
 ## Getting started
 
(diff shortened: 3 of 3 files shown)


yaacov:
@nirs please review

yaacov:
@sleviim please review

didib (approved):
LGTM. I'd also add that `./rose-server` and `./rose-client` do not return, but continue running - so you need to run them in separate shells, each with its own `pipenv shell`. In my team I suggested using `tmux`, and the students found it useful and easy (just the bare basics, of course - Ctrl-c, Ctrl-n, Ctrl-w, Ctrl-NUMBER).

yaacov:
@didib thanks, added a section about tmux

.gitignore
```diff
@@ -16,3 +16,6 @@ classroom/credentials.json
 .vscode/launch.json
 classroom/*.csv
 classroom/token.pickle
+
+#vscode
+.vscode
```
nirs, line 21: And this belongs to the "# Editors" section above.

Makefile
```diff
@@ -1,9 +1,13 @@
 init: Pipfile
+	pip install pipenv --user
+	pipenv install --python 3
```
nirs, line 3: Installing stuff with pip is bad idea, even with --user,
but if we want to use it a better way is:

    python3 -m pip ...

This way it works correctly for the used python without extra work.

Why not `dnf install pipenv`?

Makefile
```diff
@@ -1,9 +1,13 @@
...
+
+dev-init: Pipfile
 	pip install --user pipenv
-	pipenv install
+	pipenv install --python 3 --dev
```
nirs, line 7: We support only python 3 now - why do we need to specify --python 3?
Maybe you install pipenv in python2.7? If you install the pipenv package
"pipenv install" just works.

README.md
```diff
@@ -62,17 +62,17 @@ Or with pip3:
...
 
-
 ## Getting started
 
 The following commands should be performed only once; after creating the environment you will be connecting to the same environment each time you open a new session.
 
 Use pipenv to create a virtual environment and to install the rest of the dependencies:
 
-    pipenv --python /usr/local/bin/python3.7 install
+    pipenv --python 3 install
```
nirs, line 71: Since we support only python 3 now, we can simplify this to:

    pipenv install

README.md
```diff
@@ -122,6 +122,16 @@ so you can view the game in you local browser:
...
 After starting the server (as mentioned above), open a browser at http://127.0.0.1:8880/ to view and control the game. 
 
+### Openning a firewall port
+
+You can also open ports 8880 and 8888 on the remote VM running the server, and browse from a local machine
+in case port 8880 or 8888 are blocked by [firewalld](https://firewalld.org/):
+
+    DEFAULT_ZONE=$(sudo firewall-cmd --get-default-zone)
+    sudo firewall-cmd --zone=$DEFAULT_ZONE --add-port=8880/tcp --permanent
+    sudo firewall-cmd --zone=$DEFAULT_ZONE --add-port=8888/tcp --permanent
```
nirs, line 132: --zone is not needed, we can simplify.

nirs (commented):
Nice! but we can simplify now since we support only python 3 and pipenv is recent enough in current Fedora.

yaacov:
@nirs thanks, please re-review

nirs (approved):
Thanks @yaacov!

PR #450, opened 2022-07-14
README cleanups

nirs:


yaacov:
LGTM

i see I missed one `python3.7` instance :-)
the tmux commands table is much more readable now :+1:

PR #455, opened 2023-03-26
Fix flake error

sleviim:
./classroom/connect_service.py:40:11: E275 missing whitespace after keyword
    return(service)
          ^
1     E275 missing whitespace after keyword

### DIFF
--- a/classroom/connect_service.py
+++ b/classroom/connect_service.py
@@ -37,4 +37,4 @@ def create_service():
 
     service = build('classroom', 'v1', credentials=creds,
                     cache_discovery=False)
-    return (service)
+    return service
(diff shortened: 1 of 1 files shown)


classroom/connect_service.py
```diff
@@ -37,4 +37,4 @@ def create_service():
 
     service = build('classroom', 'v1', credentials=creds,
                     cache_discovery=False)
-    return(service)
+    return (service)
```
nirs, line 40: What is the purpose of the parenthesis?

This should be:

    return service

`return` is not a function.

nirs (approved):

PR #459, opened 2023-06-04
README: Fix and add some tmux commands

uril:
The default prefix-key is Ctrl-b

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,11 +190,14 @@ shell. `tmux` may be useful in this case.
 
 Example `tmux` commands:
 
-| Command | Description              |
-|---------|--------------------------|
-| Ctrl+c  | Create a new window      |
-| Ctrl+n  | Toggle to next window    |
-| Ctrl+w  | List open windows        |
+| Command    | Description                |
+|------------|----------------------------|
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -190,11 +190,14 @@ shell. `tmux` may be useful in this case.
...
 Example `tmux` commands:
 
-| Command | Description              |
-|---------|--------------------------|
-| Ctrl+c  | Create a new window      |
-| Ctrl+n  | Toggle to next window    |
-| Ctrl+w  | List open windows        |
+| Command   | Description               |
+|-----------|---------------------------|
+| Ctrl-b+c  | Create a new window       |
```
sleviim, line 195: I think that using - and + in the same context is a bit confusing.
Should it be ctrl+b+c?

uril: I see what you mean - it may be a bit confusing.
However, note that the sequence is ctrl-b (both control and b keys are pressed)
+ the following char (key is pressed after releasing both control and b keys).
Maybe leave a space to hint that:  Ctrl+b + c ?

sleviim: LGTM

PR #472, opened 2023-08-10
Print game stats to console

orilc:
At the end of the game we would like to see details about the each driver's runs, like how many penguins were collected out of possible number of penguins, number of jumps and breaks that scored points, number of collisions with obstacles.

These details are printed to the console.

Example:
```
2023-08-10 13:30:59,286 INFO    [game] Stats:
 # |            Name | Score |  Penguins |    Cracks |     Water | Collisions |
   |                 |       | Collected | Collected | Collected |            |
___________________________
 1 |          Group5 |   673 |    6 / 10 |         3 |         2 |          1 |
 2 |          Group4 |   350 |    0 / 10 |         0 |         0 |         13 | (game:123)
```

### DIFF
--- a/rose/server/game.py
+++ b/rose/server/game.py
@@ -107,13 +107,24 @@ class Game(object):
     def print_stats(self):
         lines = ['Stats:']
         top_scorers = sorted(six.itervalues(self.players), reverse=True)
+        line = '%2s | %15s | %5s | %9s | %9s | %9s | %10s |' % \
+               ('#', 'Name', 'Score', 'Penguins', 'Cracks', 'Water', 'Collisions')
+        lines.append(line)
+        line = '%2s | %15s | %5s | %9s | %9s | %9s | %10s |' % \
+               ('', '', '', 'Collected', 'Collected', 'Collected', '')
+        lines.append(line)
+        line = '_______________________________________________________________________________'
+        lines.append(line)
         for i, p in enumerate(top_scorers):
-            line = '%d  %10s  row:%d  score:%d' % (i + 1, p.name, p.y, p.score)
+            p_collected = '%d / %d' % (p.penguin_collected, self.track.penguin_counter)
+            line = '%2d | %15s | %5d | %9s | %9s | %9s | %10s |' % \
+                   (i + 1, p.name, p.score, p_collected, p.crack_collected, p.water_collected, p.collision_count)
             lines.append(line)
         log.info("%s", os.linesep.join(lines))
 
     def loop(self):
         self.track.update()
+        self.track.searching_penguin()
         score.process(self.players, self.track)
         if self.timeleft > 0:
             self.update_clients()
--- a/rose/server/player.py
+++ b/rose/server/player.py
@@ -12,6 +12,10 @@ class Player(object):
         self.action = None
         self.response_time = None
         self.score = None
+        self.penguin_collected = 0
+        self.water_collected = 0
+        self.crack_collected = 0
+        self.collision_count = 0
         self.reset()
 
     # Game state interface
@@ -25,6 +29,10 @@ class Player(object):
         self.action = actions.NONE
         self.response_time = 1.0
         self.score = 0
+        self.penguin_collected = 0
+        self.water_collected = 0
+        self.crack_collected = 0
+        self.collision_count = 0
 
     def in_lane(self):
         min_x = self.lane * config.cells_per_player
--- a/rose/server/track.py
+++ b/rose/server/track.py
@@ -6,6 +6,11 @@ class Track(object):
 
     def __init__(self):
         self._matrix = None
+        # object counters
+        self.penguin_counter = 0
+        self.crack_counter = 0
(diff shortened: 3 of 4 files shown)


orilc:
@sleviim Please approve :)

nirs:
> ```
> 2023-08-10 13:30:59,286 INFO    [game] Stats:
>  # |            Name | Score |  Penguins |    Cracks |     Water | Collisions |
>    |                 |       | Collected | Collected | Collected |            |
> ___________________________
>  1 |          Group5 |   673 |    6 / 10 |         3 |         2 |          1 |
>  2 |          Group4 |   350 |    0 / 10 |         0 |         0 |         13 | (game:123)
> ```

Thanks for sharing the example. I don't think it is very clear as is, and the code to
created is not clear. What if we keep the stats in a dict, and print the dict in simpler
yaml format instead of a table?

```
$ cat stats.py 
import sys
import yaml

stats = {
    "stats": {
        "obstacles": {
            "penguin": 10,
            "crack": 7,
            "water": 4,
        },
        "players": {
            "gropup4": {
                "penguin": 6,
                "crack": 3,
                "water": 2,
                "collision": 1,
            },
            "gropup5": {
                "penguin": 0,
                "crack": 0,
                "water": 0,
                "collision": 13,
            },
        },
    },
}

yaml.dump(stats, sys.stdout, sort_keys=False)
```

Example output:
```
$ python3 stats.py 
stats:
  obstacles:
    penguin: 10
    crack: 7
    water: 4
  players:
    gropup4:
      penguin: 6
      crack: 3
      water: 2
      collision: 1
    gropup5:
      penguin: 0
      crack: 0
      water: 0
      collision: 13
```

The advantage is keeping stats in an easy to use way inside the program, making the output machine readable so it is easy to consume by other programs, and having no code to maintain for formatting the results.

rose/server/player.py
```diff
@@ -12,6 +12,10 @@ def __init__(self, name, car, lane):
         self.action = None
         self.response_time = None
         self.score = None
+        self.penguin_collected = 0
+        self.water_collected = 0
+        self.crack_collected = 0
+        self.collision_count = 0
```
yaacov, line 18:
```suggestion
        self.penguin_collected = None
        self.water_collected = None
        self.crack_collected = None
        self.collision_count = None
```

the convention is to init to None and then use reset to set to 0

rose/server/track.py
```diff
@@ -6,6 +6,11 @@ class Track(object):
 
     def __init__(self):
         self._matrix = None
+        # object counters
+        self.penguin_counter = 0
+        self.crack_counter = 0
+        self.water_counter = 0
```
yaacov, line 12: this are never counted, need to add counters or remove them

rose/server/track.py
```diff
@@ -15,6 +20,11 @@ def update(self):
         self._matrix.pop()
         self._matrix.insert(0, self._generate_row())
 
+    # looking for all the penguins in the track
+    def searching_penguin(self):
+        if obstacles.PENGUIN in self._matrix[0]:
+            self.penguin_counter += 1
```
yaacov, line 26: better to count at the  `_generate_row` method

rose/server/track.py
```diff
@@ -39,6 +49,7 @@ def clear(self, x, y):
         self._matrix[y][x] = obstacles.NONE
 
     def reset(self):
+        self.penguin_counter = 0
```
yaacov, line 52: need to reset all the other counters too

rose/server/track.py
yaacov: better set to None, and set to zero in the reset method.

rose/server/game.py
```diff
@@ -107,13 +107,24 @@ def drive_player(self, name, info):
...
-            line = '%d  %10s  row:%d  score:%d' % (i + 1, p.name, p.y, p.score)
+            p_collected = '%d / %d' % (p.penguin_collected, self.track.penguin_counter)
+            line = '%2d | %15s | %5d | %9s | %9s | %9s | %10s |' % \
+                   (i + 1, p.name, p.score, p_collected, p.crack_collected, p.water_collected, p.collision_count)
             lines.append(line)
         log.info("%s", os.linesep.join(lines))
 
     def loop(self):
         self.track.update()
+        self.track.searching_penguin()
```
yaacov, line 127: if we count obstacles in the generate method, we don't need this call

rose/server/game.py
```diff
@@ -107,13 +107,24 @@ def drive_player(self, name, info):
     def print_stats(self):
         lines = ['Stats:']
         top_scorers = sorted(six.itervalues(self.players), reverse=True)
+        line = '%2s | %15s | %5s | %9s | %9s | %9s | %10s |' % \
```
yaacov, line 110: I agree with https://github.com/RedHat-Israel/ROSE/pull/472#issuecomment-1694720012, a stats struct and yaml dump makes more sense to me too.

PR #476, opened 2023-08-21
Use positive logic when calculating score

yaacov:
Issue:
In our score code we use refactored logic that does not explicitly follow the documented rules.

For example, when calculating the players score we use negative logic, e.g. `player.action != actions.BRAKE`  , while in the documentation we define the rules using positive language, e.g. ` if you want to continue forward. return actions.RIGHT or actions.LEFT if you like to bypass the obstacle. `

There are reasons why you might want to follow the document's logic explicitly rather than refactoring it:

- Clarity and Traceability: If someone else (or even you at a later time) refers back to the original documentation to understand or validate the code's behavior, having the logic implemented as closely as possible to the documented rules makes it easier to follow and validate.

- Avoid Introduction of Bugs: Sometimes refactoring can introduce unintended side effects, especially when the original rules have been tested and validated extensively.

- Maintenance and Future Updates: If the documentation is updated in the future, it's much easier to identify what parts of the code need to change if the code follows the structure of the documentation.

- Communication with Stakeholders: If you're communicating with non-technical stakeholders who are familiar with the documented rules but not with the nuances of code, it helps if your code closely mirrors the documentation. This can make discussions and explanations smoother.

In our case, ROSE is intended as educational code for students learning to code, it makes sense to write the score code to explicitly follow the documented rules.

What are the updates proposed by this PR:

- Use positive logic to follow the documentation, e.g. replace the `!=` with `==` and `not in` with `in`
- Use `score_pickup` explicitly, instead of using `score_move_forward`
- When moving backword use  ``score_move_backward instead of adding `score_move_forward`
 and then reducing `score_move_backward` twice.
- Move action clean up to the end of the method
- Remove six

Possible bug fixes:
- Current code assume score_move_backward and score_move_forward are identical, if they will change current code logic will break.
- Currnet code assumes score_move_forward and score_pickup are identical, if they change current code will break.

Ref:
https://github.com/RedHat-Israel/ROSE/tree/master/examples

### DIFF
--- a/rose/server/score.py
+++ b/rose/server/score.py
@@ -27,48 +27,64 @@ def process(players, track):
             if player.x < config.matrix_width - 1:
                 player.x += 1
 
-    # Now handle obstacles, preferring players in their own lane.
+    # Proccess the players by order, first the ones in lane and then
+    # the ones out of lane, this ensure the car in lane will have
+    # priority when picking pinguins and in case of collisions.
 
     sorted_players = sorted(six.itervalues(players),
                             key=lambda p: 0 if p.in_lane() else 1)
     positions = set()
 
+    # Now handle obstacles, preferring players in their own lane.
+
     for player in sorted_players:
-        player.score += config.score_move_forward
         obstacle = track.get(player.x, player.y)
-        if obstacle == obstacles.CRACK:
-            if player.action != actions.JUMP:
-                track.clear(player.x, player.y)
-                player.y += 1
-                player.score += config.score_move_backward * 2
-            else:
-                player.score += config.score_jump
+
+        if obstacle == obstacles.NONE:
+            # Move forward, leaving the obstacle on the track.
+            player.score += config.score_move_forward
+
         elif obstacle in (obstacles.TRASH,
                           obstacles.BIKE,
                           obstacles.BARRIER):
-            if player.action not in (actions.LEFT, actions.RIGHT):
-                track.clear(player.x, player.y)
-                player.y += 1
-                player.score += config.score_move_backward * 2
-        elif obstacle == obstacles.WATER:
-            if player.action != actions.BRAKE:
-                track.clear(player.x, player.y)
-                player.y += 1
-                player.score += config.score_move_backward * 2
+            # Move back consuming the obstacle.
+            track.clear(player.x, player.y)
+            player.y += 1
+            player.score += config.score_move_backward
+
+        elif obstacle == obstacles.CRACK:
+            if player.action == actions.JUMP:
+                # Move forward leaving the obstacle on the track
+                player.score += config.score_move_forward + config.score_jump
             else:
-                player.score += config.score_brake
+                # Move back consuming the obstacle.
+                track.clear(player.x, player.y)
+                player.y += 1
+                player.score += config.score_move_backward
+
+        elif obstacle == obstacles.WATER:
+            if player.action == actions.BRAKE:
+                # Move forward leaving the obstacle on the track
+                player.score += config.score_move_forward + config.score_brake
+            else:
+                # Move back consuming the obstacle.
+                track.clear(player.x, player.y)
+                player.y += 1
+                player.score += config.score_move_backward
+
         elif obstacle == obstacles.PENGUIN:
             if player.action == actions.PICKUP:
+                # Move forward and collect an aquatic bird
                 track.clear(player.x, player.y)
+                player.score += config.score_move_forward + config.score_pickup
+            else:
+                # Move forward leaving the obstacle on the track
                 player.score += config.score_move_forward
 
         # Here we can end the game when player gets out of
         # the track bounds. For now, just keep the player at the same
         # location.
         player.y = min(config.matrix_height - 1, max(2, player.y))
-
-        # Finally forget action
-        player.action = actions.NONE
 
         # Fix up collisions
 
@@ -83,9 +99,12 @@ def process(players, track):
             elif player.x < config.matrix_width - 1:
                 player.x += 1
 
+        # Finally forget action
+        player.action = actions.NONE
+
+        positions.add((player.x, player.y))
+
         log.info('process_actions: name=%s lane=%d pos=%d,%d score=%d '
                  'response_time=%0.6f',
                  player.name, player.lane, player.x, player.y, player.score,
                  player.response_time)
-
-        positions.add((player.x, player.y))
--- a/rose/common/config.py
+++ b/rose/common/config.py
@@ -39,6 +39,7 @@ cells_per_player = matrix_width // max_players
 
 score_move_forward = 10
 score_move_backward = -10
+score_pickup = 10
 score_jump = 5
 score_brake = 4
 


yaacov:

rose-admin
```diff
@@ -4,12 +4,7 @@ import sys
 import os
 import logging
 
-import six
-
-if six.PY2:
-    import xmlrpclib
-else:
-    from xmlrpc import client as xmlrpclib
+from xmlrpc import client as xmlrpclib
```
nirs, line 7: This is unrelated change (remove python 2 leftovers) best done in a separate commit.

rose/client/game.py
```diff
@@ -1,8 +1,6 @@
 import logging
 import time
 
-import six
-
```
nirs, line 5: Same unrelated python 2 cleanups.

rose/server/score.py
```diff
@@ -19,62 +17,72 @@ def process(players, track):
...
+        obstacle = track.get(player.x, player.y)
 
-    sorted_players = sorted(six.itervalues(players),
-                            key=lambda p: 0 if p.in_lane() else 1)
-    positions = set()
+        if obstacle == obstacles.NONE:
+            # Move forlward
+            player.score += config.score_move_forward
 
-    for player in sorted_players:
```
nirs, line 36: I don't see how the change is correct when you don't process players in order.

The purpose of processing in order is to give priority to some players, so they
can acquire the next position before the other players. In the past we use the 
player response time to give priority to faster players but this was removed since
players do not use the same machines on the same network.

rose/server/score.py
```diff
@@ -19,62 +17,72 @@ def process(players, track):
...
-        player.action = actions.NONE
+        # Filter for players with the same x and y
+        all_collisions = filter(lambda p:
+                                p.name != player.name and
+                                p.x == player.x and
+                                p.y == player.y, players.values())
+        next_collision = next(all_collisions, None)
 
-        # Fix up collisions
+        # if we have a collision and we are not in our lane, try to move back
```
nirs, line 81: This depends on order of players, this is why we process players in some order. The previous player got this position so the next players need to find another position.

nirs (commented):
Using positive logic and avoiding assumptions so we can change the rules easily in the future sounds like a good idea. I'm not sure about
the change to the player loop, and I don't think that avoiding the sorting is needed since our N is tiny.

Can you split out the unrelated python 2 changes and the positive logic changes to different commits?

We can consider changing the player loop later if you want.

rose-admin
yaacov: Thanks, created a python2 cleanup PR #478

yaacov: Removed the python2 cleanup code in favor or #478

rose/server/score.py
yaacov: > The purpose of processing in order is to give priority to some players

this is not documented in https://github.com/RedHat-Israel/ROSE/tree/master/examples

the documented rule is:

> NOTE : If you are driving out of your lane and have a collision with the lane owner, you will punished by -10 points and be moved to other free place.

The purpose of this PR is to align the code with the documentation,
So the change is correct to follow the documentation, it removes the undocumented logic:

change from checking the priority by sorting using undocumented criteria, 
to the explicitly documented rule of giving priority to the car that is in it's lane:
"If you are driving out of your lane"

When we update the documentation and change the collision rules (i don't think we should) we can update the code accordingly

rose/server/score.py
yaacov: The documentation say:

>NOTE : If you are driving out of your lane and have a collision with the lane owner, you will punished by -10 points and be moved to other free place.

so this is the currect behavior according to the documentation

Ref:
https://github.com/RedHat-Israel/ROSE/tree/master/examples

nirs: The documentation does not describe exactly the semantics. Handling collisions is much more complicated than what we describe in the docs.

We process the gate state in 3 steps:
1. Change the location, needed only for RIGHT or LEFT actions. We can see this as our physics engine :-). This is done in no particular order.
2. Handle obstacles - this must be done in order, since when a player pick up or hit an obstacle, we remove the obstacle from the track. The next player will not hit this obstacle since it does not exist at this point.
3. Fix up collisions by moving the player to another position if another player got this position before.

You can see this more clearly using the new debug logs I added in #479.

We can change the way this works, but we can also change the documentation to describe better how collision are handled.

You propose many change in this pr:
> Use positive logic to follow the documentation, e.g. replace the != with == and not in with in

This sounds like good change for entire pr.

> When testing collisions xplicitly test for in_lane instead of using the implicit sort technique

I'm not sure what implicit about the current code, and I don't see how the suggested code is better and how it handles correctly obstacles, so only the first player pick/hit the obstacle.

> Use score_pickup explicitly, instead of using score_move_forward

Sounds good, simple change that should be in its own commit.

> When moving backword use ``score_move_backward instead of adding score_move_forward
and then reducing `score_move_backward` twice.

Sounds good for its own commit - the current code is confusing. It is better if we change the score in single place at the processing loop.

> Move action clean up to the end of the method

Sounds good for its own commit. Maybe this should happen at the point we process the action - once we processed the action it can be set to None so we cannot process the same action twice.

I suggest to focus on the simple changes that cannot break anything first to get rid of them and make the code cleaner before we try to handle the bigger changes, where we are not sure if we need to change the code or the docs.

rose/server/score.py
```diff
@@ -27,54 +27,64 @@ def process(players, track):
...
-                player.y += 1
-                player.score += config.score_move_backward * 2
-            else:
-                player.score += config.score_jump
         elif obstacle in (obstacles.TRASH,
                           obstacles.BIKE,
                           obstacles.BARRIER):
-            if player.action not in (actions.LEFT, actions.RIGHT):
+            # Move back
+            track.clear(player.x, player.y)
```
nirs, line 40: This consumes the obstacle - since you don't sort the players, the player driving not it its lane can consume the obstacle before the other player. This does not look like the right direction.

rose/server/score.py
```diff
@@ -27,54 +27,64 @@ def process(players, track):
...
 
-    sorted_players = sorted(six.itervalues(players),
-                            key=lambda p: 0 if p.in_lane() else 1)
-    positions = set()
+        if obstacle == obstacles.NONE:
+            # Move forlward
+            player.score += config.score_move_forward
 
-    for player in sorted_players:
-        player.score += config.score_move_forward
```
nirs, line 37: Removing this and replacing it with score for every match is good.

rose/server/score.py
```diff
@@ -27,54 +27,64 @@ def process(players, track):
...
+            # Move back
+            track.clear(player.x, player.y)
+            player.y += 1
+            player.score += config.score_move_backward
+
+        elif obstacle == obstacles.CRACK:
+            if player.action == actions.JUMP:
+                # Move forlward
+                player.score += config.score_move_forward
+                player.score += config.score_jump
```
nirs, line 48: We can add the score with one line:

     player.score += config.score_move_forward + config.score_jump

rose/server/score.py
```diff
@@ -27,54 +27,64 @@ def process(players, track):
...
+            else:
+                # Move back
                 track.clear(player.x, player.y)
                 player.y += 1
-                player.score += config.score_move_backward * 2
+                player.score += config.score_move_backward
```
nirs, line 53: Nice to get rid of this unclear `* 2`

rose/server/score.py
```diff
@@ -27,54 +27,64 @@ def process(players, track):
...
                 player.score += config.score_move_forward
+                player.score += config.score_pickup
+            else:
+                # Move forlward no reward
+                player.score += config.score_move_forward
 
-        # Here we can end the game when player gets out of
-        # the track bounds. For now, just keep the player at the same
-        # location.
-        player.y = min(config.matrix_height - 1, max(2, player.y))
```
nirs, line 68: You moved this later, but should be done before we handle collisions. If a player was move back out of the track by hitting an obstacle, we want to move it back to the track. We can do this at the point we change player.y, maybe:

    player.move_backward()

This can do the checks for out-of-track so we don't need to check in the processing loop.

Which is much more clear than:

    player.y += 1

(Working with the inverted coordinates is horrible)

rose/server/score.py
```diff
@@ -27,54 +27,64 @@ def process(players, track):
...
-
-        # Finally forget action
-        player.action = actions.NONE
+        # Filter for players with the same x and y
+        all_collisions = filter(lambda p:
+                                p.name != player.name and
+                                p.x == player.x and
+                                p.y == player.y, six.itervalues(players))
```
nirs, line 80: I don't understand this code is doing - mostly likely nobody else will.

rose/server/score.py
yaacov: reverted to sorted players again - so this will work now

rose/server/score.py
yaacov: makes sense to me too, moved back to before the collision check

rose/server/score.py
yaacov: done

rose/server/score.py
yaacov: removed in favor or reverting back to the sorting method

rose/server/score.py
yaacov: > I suggest to focus on the simple changes that cannot break anything first to get rid of them and make the code cleaner before we try to handle the bigger changes, where we are not sure if we need to change the code or the docs.

ok, reverted back to using the sorting method, and added a comment explaining it's logic

rose/server/score.py
```diff
@@ -27,49 +27,65 @@ def process(players, track):
...
-        if obstacle == obstacles.CRACK:
-            if player.action != actions.JUMP:
-                track.clear(player.x, player.y)
-                player.y += 1
-                player.score += config.score_move_backward * 2
-            else:
-                player.score += config.score_jump
+
+        if obstacle == obstacles.NONE:
+            # Move forlward
```
nirs, line 44: typo: forlward

Maybe make the comment more clear about whats going on:

    # Move forward, leaving the obstacle on the track.

rose/server/score.py
```diff
@@ -27,49 +27,65 @@ def process(players, track):
...
+            player.score += config.score_move_forward
+
         elif obstacle in (obstacles.TRASH,
                           obstacles.BIKE,
                           obstacles.BARRIER):
-            if player.action not in (actions.LEFT, actions.RIGHT):
+            # Move back
```
nirs, line 50: Maybe:

    # Move back consuming the obstacle.

rose/server/score.py
```diff
@@ -27,49 +27,65 @@ def process(players, track):
...
+            track.clear(player.x, player.y)
+            player.y += 1
+            player.score += config.score_move_backward
+
+        elif obstacle == obstacles.CRACK:
+            if player.action == actions.JUMP:
+                # Move forlward
```
nirs, line 57: Maybe:

    # Move forward leaving the obstacle on the track

rose/server/score.py
```diff
@@ -83,9 +99,11 @@ def process(players, track):
             elif player.x < config.matrix_width - 1:
                 player.x += 1
 
+        # Finally forget action
```
nirs, line 102: This comment looks like a block comment for the next 2 lines, but is is a comment only for the first line. Adding a blank line between the next lines will make this more clear.

nirs (commented):
Changes look good but comments can be improved.

rose/server/score.py
yaacov: done

rose/server/score.py
yaacov: done

rose/server/score.py
yaacov: done

rose/server/score.py
yaacov: done

nirs (approved):
Thanks!

PR #477, opened 2023-08-21
Explictly test for players car in lane

yaacov:
Issue:
When testing if a car is in lane, we do an implicit test, calculating the start and end x value of the player lane, it's better to use explicit calculation to get the lane directly from car position.

### DIFF
--- a/rose/server/player.py
+++ b/rose/server/player.py
@@ -27,9 +27,8 @@ class Player(object):
         self.score = 0
 
     def in_lane(self):
-        min_x = self.lane * config.cells_per_player
-        next_x = min_x + config.cells_per_player
-        return min_x <= self.x < next_x
+        current_lane = self.x // config.cells_per_player
+        return current_lane == self.lane
 
     def __cmp__(self, other):
         x = self.score
--- a/rose/server/player.py
+++ b/rose/server/player.py
@@ -4,6 +4,25 @@ from rose.common import actions, config
 class Player(object):
 
     def __init__(self, name, car, lane):
+        """
+        Creates a new Driver object.
+        Args:
+            name (str): The unique name of the driver for display.
+            car (int): A number from 0 to 3, representing the car type.
+            lane (int): The initial lane number, between 0 and the
+                            maximum lane.
+
+        Attributes:
+            x (int, optional): The X-coordinate of the driver.
+                                Starts in the middle of the lane.
+            y (int, optional): The Y-coordinate of the driver.
+                                Starts at 2/3 of the screen height.
+            action (str, optional): The current driver action. Defaults
+                                        to 'none'.
+            response_time (float, optional): The duration the driver takes
+                                                to react. Starts as None.
+            score (int, optional): The driver's current score. Begins at 0.
+        """
         self.name = name
         self.car = car
         self.lane = lane
@@ -23,7 +42,7 @@ class Player(object):
         self.x = self.lane * config.cells_per_player + 1  # | |0| | |1 | |
         self.y = config.matrix_height // 3 * 2             # 1/3 of track
         self.action = actions.NONE
-        self.response_time = 1.0
+        self.response_time = None
         self.score = 0
 
     def in_lane(self):
--- a/rose/server/player_test.py
+++ b/rose/server/player_test.py
@@ -1,15 +1,80 @@
+from rose.common import actions, config
 from . import player
 
 
-def test_in_lane():
-    p = player.Player("A", car=0, lane=0)
-    for x in (0, 1, 2):
-        p.x = x
-        assert p.in_lane()
+def test_player_initialization():
+    player1 = player.Player("John", 1, 1)
+
+    assert player1.name == "John"
+    assert player1.car == 1
+    assert player1.lane == 1
+    # player.Player x value should be middle of it's lane
+    assert player1.x == player1.lane * config.cells_per_player + 1
+    assert player1.y == config.matrix_height // 3 * 2
+    assert player1.action == actions.NONE
+    assert player1.response_time is None
+    assert player1.score == 0
 
 
-def test_not_in_lane():
-    p = player.Player("A", car=0, lane=1)
-    for x in (0, 1, 2):
-        p.x = x
-        assert not p.in_lane()
+def test_player_reset():
+    player1 = player.Player("John", 1, 1)
+
+    player1.score = 50  # Modify player to make sure reset works
+    player1.reset()
+    assert player1.x == player1.lane * config.cells_per_player + 1
+    assert player1.y == config.matrix_height // 3 * 2
+    assert player1.action == actions.NONE
+    assert player1.response_time is None
+    assert player1.score == 0
+
+
+def test_player_in_lane():
+    player1 = player.Player("John", 1, 1)
+
+    lane_start = player1.lane * config.cells_per_player
+
+    for offset in range(config.cells_per_player):
+        player1.x = lane_start + offset
+        assert player1.in_lane()
+
+
+def test_player_not_in_lane():
+    player1 = player.Player("John", 1, 1)
+
+    # Modify player's position to be out of their lane
+    other_lane = (player1.lane + 1) % config.max_players
+    other_lane_start = other_lane * config.cells_per_player
+
+    for offset in range(config.cells_per_player):
+        player1.x = other_lane_start + offset
+        assert not player1.in_lane()
+
+
+def test_player_comparison():
+    player1 = player.Player("John", 1, 1)
+    player2 = player.Player("Doe", 2, 2)
+
+    # Initial scores are the same
+    assert not player1 < player2
+    assert not player2 < player1
+
+    # Modify scores
+    player1.score = 10
+    player2.score = 20
+
+    assert player1 < player2
+
+
+def test_player_state():
+    player1 = player.Player("John", 1, 1)
+
+    expected_state = {
+        'name': player1.name,
+        'car': player1.car,
+        'x': player1.x,
+        'y': player1.y,
+        'lane': player1.lane,
+        'score': player1.score
+    }
+
+    assert player1.state() == expected_state


yaacov:

rose/server/player.py
```diff
@@ -27,9 +27,8 @@ def reset(self):
         self.score = 0
 
     def in_lane(self):
-        min_x = self.lane * config.cells_per_player
-        next_x = min_x + config.cells_per_player
-        return min_x <= self.x < next_x
+        current_lane = self.x // config.cells_per_player
```
nirs, line 30: Not very clear why `x // cells_per_player` works but it is shorter so why not. To make sure we don't get this wrong, adding couple of trivial tests can be nice.

nirs (approved):

nirs:
> Issue: When testing if a car is in lane, we do an implicit test, calculating the start and end x value of the player lane, it's better to use explicit calculation to get the lane directly from car position.

Not clear to me why the current code is implicit and the new code is explicit. I think both
are the same in this aspect.

rose/server/player.py
yaacov: thanks:
added some tests

rose/server/player_test.py
```diff
@@ -1,15 +1,88 @@
...
+
+
+def test_player_state(player):
+    state = player.state()
+    assert state['name'] == player.name
+    assert state['car'] == player.car
+    assert state['x'] == player.x
+    assert state['y'] == player.y
+    assert state['lane'] == player.lane
+    assert state['score'] == player.score
```
nirs, line 88: This can be more clear by comparing to literal dict. Do we have reason not to compare the state to a dict (e.g. time values)?

rose/server/player_test.py
```diff
@@ -1,15 +1,88 @@
...
+])
+def test_player_in_lane_various_positions(player, lane, x, expected):
+    player.lane = lane
+    player.x = x
+    assert player.in_lane() == expected
+
+
+def test_player_comparison():
+    player1 = Player("John", "car1", 1)
+    player2 = Player("Doe", "car2", 2)
```
nirs, line 68: Good example why the fixture is not needed, this test is more clear than the previous tests since the code creating the players is right here.

rose/server/player_test.py
```diff
@@ -1,15 +1,88 @@
...
 
-def test_in_lane():
-    p = player.Player("A", car=0, lane=0)
-    for x in (0, 1, 2):
-        p.x = x
-        assert p.in_lane()
+# Sample Test Setup
+@pytest.fixture
+def player():
+    return Player("John", "car1", 1)
```
nirs, line 9: I don't think the fixture helps here - we don't have any complex state to manage, and the tests will be more clear if we create the player right in the test as you do in some of the tests.

A fixture can be useful to test with different configurations, like different values of cells_per_player.

rose/server/player_test.py
```diff
@@ -1,15 +1,88 @@
...
-        p.x = x
-        assert not p.in_lane()
+def test_player_initialization(player):
+    assert player.name == "John"
+    assert player.car == "car1"
+    assert player.lane == 1
+    assert player.x is not None  # This checks that x is initialized
+    assert player.y == config.matrix_height // 3 * 2
+    assert player.action == actions.NONE
+    assert player.response_time == 1.0
```
nirs, line 19: Strange that we have a value for new player that never had any response. Good that you added the test to reveal this issue.

rose/server/player_test.py
```diff
@@ -1,15 +1,88 @@
...
-def test_not_in_lane():
-    p = player.Player("A", car=0, lane=1)
-    for x in (0, 1, 2):
-        p.x = x
-        assert not p.in_lane()
+def test_player_initialization(player):
+    assert player.name == "John"
+    assert player.car == "car1"
+    assert player.lane == 1
+    assert player.x is not None  # This checks that x is initialized
```
nirs, line 16: Better to test that you cannot create a player without a x value (e.g. it can raise).

rose/server/player_test.py
```diff
@@ -1,15 +1,88 @@
...
+    assert player.x == player.lane * config.cells_per_player + 1
+    assert player.y == config.matrix_height // 3 * 2
+    assert player.action == actions.NONE
+    assert player.response_time == 1.0
+    assert player.score == 0
+
+
+def test_player_in_lane_true(player):
+    # This test assumes the player is in their lane after reset.
+    # So, we test the default state after initialization.
```
nirs, line 35: This comment is needed since the player fixture hides the code to create the player.

rose/server/player_test.py
```diff
@@ -1,15 +1,88 @@
...
+    assert player.in_lane() is True
+
+
+def test_player_in_lane_false(player):
```
nirs, line 39: maybe test_player_not_in_lane()?

rose/server/player_test.py
```diff
@@ -1,15 +1,88 @@
...
+    (1, 1, False),  # Lane 1, x=1
+    (1, 2, False),  # Lane 1, x=2
+    (1, 3, True),   # Lane 1, x=3
+    (1, 4, True),   # Lane 1, x=4
+    (1, 5, True)    # Lane 1, x=5
+])
+def test_player_in_lane_various_positions(player, lane, x, expected):
+    player.lane = lane
+    player.x = x
+    assert player.in_lane() == expected
```
nirs, line 63: Why 3 tests for the same thing? we have test_player_in_lane(), test_player_in_lane_false() and this one.

This tests seems to depend on the configuration cells_per_lane - if we change it the test will break.

rose/server/player_test.py
```diff
@@ -1,15 +1,88 @@
...
+
+def test_player_in_lane_false(player):
+    # Modify player's position to be out of their lane
+    other_lane = (player.lane + 1) % config.max_players
+    player.x = other_lane * config.cells_per_player + 1
+    assert player.in_lane() is False
+
+
+@pytest.mark.parametrize("lane,x,expected", [
+    (0, 0, True),   # Lane 0, x=0
```
nirs, line 47: The comment tell use the code is not clear enough. This is a common thing when using parameters. Writing simpler with a loop avoid this issue.

But we also make this more clear using a dict:

    dict(lane=0, x=0, in_lane=True),

nirs (commented):
Thanks the tests!

rose/server/player_test.py
yaacov: do you suggest to add a specific test to test that x is not None after creation ?

rose/server/player_test.py
yaacov: removed the fixture, we create player a lot of time, so it can go both ways, a matter of taste :-)

rose/server/player_test.py
yaacov: removed the fixture, IMHO it is a matter of taste because the fixture prevent code repetitions not just for different configurations.

rose/server/player_test.py
yaacov: :-), changed the init method to start with None, and updated the test also

rose/server/player_test.py
yaacov: done

rose/server/player_test.py
yaacov: removed the duplicated test, it was an over kill, but this PR is about the in_lane method :-)

rose/server/player_test.py
yaacov: removed the duplicated test, it was an over kill, but this PR is about the in_lane method :-)

rose/server/player_test.py
yaacov: No :-)
just speed of making the tests :runner:

rose/server/player_test.py
nirs: I suggest to change the player so we cannot create a player in invalid state if this is not the case already. And in this case we probably know what is x so we don't need to check for `is not None` or add a comment.

rose/server/player_test.py
```diff
@@ -1,15 +1,74 @@
...
+from rose.server.player import Player
 
 
-def test_in_lane():
-    p = player.Player("A", car=0, lane=0)
-    for x in (0, 1, 2):
-        p.x = x
-        assert p.in_lane()
+def test_player_initialization():
+    player = Player("John", "car1", 1)
```
nirs, line 6: This is not documented, but based on rose/server/game.py, car is an integer.

rose/server/player_test.py
nirs: Looking at player we don't need to change anything, we set x based on the lane. So we can simply test that we set x to the right value:

```
 22     def reset(self):
 23         self.x = self.lane * config.cells_per_player + 1  # | |0| | |1 | |
 24         self.y = config.matrix_height // 3 * 2             # 1/3 of track
```

rose/server/player_test.py
yaacov: oyy, yes, it's an int ...

a. changed to int
b. added documentation, so it will be clear it's an int

rose/server/player_test.py
```diff
@@ -1,15 +1,74 @@
-from . import player
+from rose.common import actions, config
+from rose.server.player import Player
```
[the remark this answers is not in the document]
yaacov: ok, rules are rules, reverted the import

rose/server/player.py
```diff
@@ -23,7 +23,7 @@ def reset(self):
         self.x = self.lane * config.cells_per_player + 1  # | |0| | |1 | |
         self.y = config.matrix_height // 3 * 2             # 1/3 of track
         self.action = actions.NONE
-        self.response_time = 1.0
+        self.response_time = None
```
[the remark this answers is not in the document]
yaacov: :+1:

rose/server/player_test.py
yaacov: changed to testing that x in the middle of the lane (assuming lane width of 3 .. but the code does that too, so it's ok)

rose/server/player.py
```diff
@@ -4,6 +4,23 @@
...
+    Attributes:
+      x (int, optional): The X-coordinate of the driver.
+                         Starts in the middle of the lane.
+      y (int, optional): The Y-coordinate of the driver.
+                         Starts at 2/3 of the screen height.
+      action (str, optional): The current driver action. Defaults to 'none'.
+      response_time (float, optional): The duration the driver takes to react.
+                                       Starts as None.
+      score (int, optional): The driver's current score. Begins at 0.
+"""
```
nirs, line 23: Indentation is off, should be:

    def __init__(...):
        """
        title ...
 
        body text...
        """

rose/server/player_test.py
```diff
@@ -1,15 +1,75 @@
...
+    player1.score = 50  # Modify player to make sure reset works
+    player1.reset()
+    assert player1.x == player1.lane * config.cells_per_player + 1
+    assert player1.y == config.matrix_height // 3 * 2
+    assert player1.action == actions.NONE
+    assert player1.response_time is None
+    assert player1.score == 0
+
+
+def test_player_in_lane_true():
```
nirs, line 31: test_player_in_lane()?

rose/server/player_test.py
```diff
@@ -1,15 +1,75 @@
...
+    # So, we test the default state after initialization.
+    assert player1.in_lane() is True
+
+
+def test_player_not_in_lane():
+    player1 = player.Player("John", 1, 1)
+
+    # Modify player's position to be out of their lane
+    other_lane = (player1.lane + 1) % config.max_players
+    player1.x = other_lane * config.cells_per_player + 1
```
nirs, line 44: Would be nice to test that all 3 possible cells in the other lane are handle correctly, something like:

    base = other_lane * other_lane * config.cells_per_player
    for offset in range(config.cells_per_player):
        player.x = base + offset
        assert not player1.in_lane()

rose/server/player_test.py
```diff
@@ -1,15 +1,75 @@
...
+    assert player1.response_time is None
+    assert player1.score == 0
+
+
+def test_player_in_lane_true():
+    player1 = player.Player("John", 1, 1)
+
+    # This test assumes the player is in their lane after reset.
+    # So, we test the default state after initialization.
+    assert player1.in_lane() is True
```
nirs, line 36: This works, but more idiomatic check would be:

    assert player1.in_lane()

Checking for is True is needed only for very special cases when we want to ensure specific value (True, False) and not any true ("1", ["not empty"]) or false value ("", []).

rose/server/player.py
yaacov: wow, :man_facepalming: , thanks ! fixed

rose/server/player_test.py
yaacov: done

rose/server/player_test.py
yaacov: done :-)

rose/server/player_test.py
yaacov: nice, fixed

rose/server/player.py
```diff
@@ -4,6 +4,25 @@
...
+                x (int, optional): The X-coordinate of the driver.
+                                    Starts in the middle of the lane.
+                y (int, optional): The Y-coordinate of the driver.
+                                    Starts at 2/3 of the screen height.
+                action (str, optional): The current driver action. Defaults
+                                         to 'none'.
+                response_time (float, optional): The duration the driver takes
+                                                  to react. Starts as None.
+                score (int, optional): The driver's current score. Begins at 0.
+        """
```
nirs, line 25: Indentation is off again - the entire contents is indented too much.

It should be:

```python
    def __init__(self, name, car, lane):
        """
        Creates a new Driver object.
        Args:
            name (str): The unique name of the driver for display.
            ...

        Attributes:
            x (int, optional): The X-coordinate of the driver.
                Starts in the middle of the lane.
            ...
        """
```

See https://stackoverflow.com/questions/3898572/what-are-the-most-common-python-docstring-formats for all 4(!) docstring formats.

yaacov: fixed, 3rd time ice cream :ice_cream:

yaacov: in flake8 we trust ... need some pydoc checks ...

nirs (approved):

PR #478, opened 2023-08-26
Cleanup python2 six and imp libraries

yaacov:
Ref: https://github.com/RedHat-Israel/ROSE/pull/476/files#r1306497974

Follow up for score logic cleanup, separate PR #476 to python2 cleanup and score logic cleanup.

### DIFF
--- a/rose/client/main.py
+++ b/rose/client/main.py
@@ -1,7 +1,6 @@
 import argparse
-import imp
+import importlib
 import logging
-import os.path
 import sys
 
 from twisted.internet import reactor, protocol
@@ -73,7 +72,7 @@ class ClientFactory(protocol.ReconnectingClientFactory):
 
 def load_driver_module(file_path):
     """
-    Load the driver module from file
+    Load the driver module from the specified path.
 
     Arguments:
       file_path (str): The path to the driver module
@@ -82,16 +81,12 @@ def load_driver_module(file_path):
         Driver module (module)
 
     Raises:
-        ImportError if the module cannot be loaded
+        Exception if the module cannot be loaded
     """
-    module_dir, module_name = os.path.split(os.path.splitext(file_path)[0])
-    fp, pathname, description = imp.find_module(module_name, [module_dir])
-    try:
-        return imp.load_module(module_name, fp, pathname, description)
-    finally:
-        # Since we may exit via an exception, close fp explicitly.
-        if fp:
-            fp.close()
+    spec = importlib.util.spec_from_file_location("driver_module", file_path)
+    driver_module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(driver_module)
+    return driver_module
 
 
 def main():
@@ -109,7 +104,7 @@ def main():
 
     try:
         driver_mod = load_driver_module(args.driver_file)
-    except ImportError as e:
+    except Exception as e:
         log.error("error loading driver module %r: %s", args.driver_file, e)
         sys.exit(2)
 
--- a/rose-admin
+++ b/rose-admin
@@ -4,12 +4,7 @@ import sys
 import os
 import logging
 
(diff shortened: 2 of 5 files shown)


yaacov:

rose/client/main.py
```diff
@@ -71,27 +70,23 @@ def send_message(self, msg):
...
-      file_path (str): The path to the driver module
-
-    Returns:
-        Driver module (module)
-
-    Raises:
-        ImportError if the module cannot be loaded
+    :param driver_path: Path to the driver module.
+    :return: The loaded module.
+    :raises ImportError: If there's an issue loading the module.
```
nirs, line 79: Why use the more cryptic styntax instead of the current style? Both are valid syntax for docstrings and supported by documentation tools. Lees keep the current style.

rose/client/main.py
```diff
@@ -71,27 +70,23 @@ def send_message(self, msg):
...
-        if fp:
-            fp.close()
+        spec = importlib.util.spec_from_file_location(
+            "driver_module", driver_path)
+        driver_module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(driver_module)
+        return driver_module
+    except Exception as e:
+        raise ImportError(
+            f"Error loading driver module from path {driver_path}: {str(e)}")
```
nirs, line 89: Since we cannot handle this error, it is better not to handle it and let if propagate to the top level caller. Changing the error is also not related to removing python 2 leftovers so it is best to just replace usage of the imp modujle with the modren version and leave error handling as is. We can revisit error handling later if needed.

nirs (commented):
python 2 cleanups looks good.

rose/client/main.py
yaacov: Thanks, reverted to the original pydoc

rose/client/main.py
yaacov: yeap, makes sense to me too, remvoed

nirs (approved):

sleviim:
Just before merging, want to verify - @yaacov have you tried running the code with those changes and it went well? :)

yaacov:
> Just before merging, want to verify - @yaacov have you tried running the code with those changes and it went well? :)

Yes :-)

( note that the `.pyc` examples will not work on python3 )

screenshot-localhost_8880-2023 08 27-11_01_03

sleviim:
hmmm interesting
is there a way to run drivers codes without exposing the code (as we used the pyc for it)?

yaacov:
> hmmm interesting is there a way to run drivers codes without exposing the code (as we used the pyc for it)?

sorry for the confusion, in general, .pyc files are specific to the version of Python with which they were compiled, due to the potential changes in bytecode between versions. This is the reason why the magic number exists in the .pyc header; it helps prevent the accidental execution of a bytecode file with a different interpreter version than what it was compiled for.

to make the current` .pyc` files work you need to run them using the same version of python they where created with:

```` bash
hexdump -n 2 -e '1/2 "%04x"' score.pyc 
0d42
# 0d42 is magic for python 3.9
```
using any other version of python may or may not run the `.pyc` files currectly.

> is there a way to run drivers codes without exposing the code (as we used the pyc for it)?

a. yes, recompile them using the currently used python (you will need the sources for that)
b. using pipenv, venv or Dockerfile (if we move to containers) to pin down the python version

yaacov:
for completeness, link to python magic numbers:
https://github.com/google/pytype/blob/main/pytype/pyc/magic.py

yaacov:
side note:
if we replace the RPC model with classic HTTP client-server model, we can write clients in any language, including compiled ones (e.g. rust / go ... )
and then we will not have the python decompile `.pyc => .py` problem

nirs (approved):

PR #470, opened 2023-08-10
changed the Play button to Music button

itayfree:
Submitted by Itay Hofshy, Avia Izakovitch, Gavriel Fernanadez.

why?
While working on our driver's code, we got confused many times by the "Play" button, since it seemed that this button starts the game, but it actually starts the music. We changed its name to "Music" so that the player will not get confused anymore.

Screenshots:

PR #480, opened 2023-08-27
js,html: Change the 'Play' button into a 'Music' button

sleviim:
This patch was suggested during the Nitzanim hackathon


yaacov (approved):
lgtm,

but shouldn't we merge #470 ?

Ref https://github.com/RedHat-Israel/ROSE/pull/470/files

sleviim:
YES Thank you!!!
I didn;t find it :)

PR #481, opened 2023-08-27
Use more idiomatic docstring indentation

nirs:


yaacov (approved):
thanks, this does not count as 3rd time :-)

yaacov:
p.s.
IMHO we need a dot tile for flake8:
```
[flake8]
max-line-length = 240
```
the 80 chars rule is not fun :-)

nirs:
Nice code have short and sweet names and only few indentation levels and works fine with 80 characters.

But we can switch to black auto formatter, which accepts longer lines for the few case they are needed (e.g. long url in comments) and save the work for formatting the code.

Added #482 with example on how to integrate it.

PR #485, opened 2023-09-11
Make ROSE containers more friendly

yaacov:
**TL;DR demo**

``` bash
# Start student A driver container
podman run --rm --network host -it quay.io/rose/rose-ml-driver:latest --port 8081

# Start student B driver container
podman run --rm --network host -it quay.io/rose/rose-go-driver:latest --port 8082

# Start server container, and connect to the students drivers
podman run --rm --network host -it quay.io/rose/rose-server:latest --drivers \
     http://127.0.0.1:8081 http://127.0.0.1:8082

# Open browser on local machine at: http://127.0.0.1:8880
```

[Screencast from 2023-09-11 16-09-37.webm](https://github.com/RedHat-Israel/ROSE/assets/2181522/d86ef3f9-a16e-45f3-93b1-8949a987f8cb)

Ref:
Python machine learning driver: https://github.com/yaacov/rose-ml-driver
Go land driver: https://github.com/yaacov/rose-go-driver

**What this POC try to solve:**

a. We can not create a k8s Pod for each driver, and let them run forever.
    When a driver starts it immediately try to connect to the server and blocks other drivers from running.
b. We use python RPC protocol to comunicate between the driver and the server, this makes it hard to write
    drivers in compiled languages, making us relay on .pyc file to hide implementation code.
c. We use 2.7 oriented libraries: twisted, autobahn and pytest.

**Fix a and b:**
    Replace the RPC calls with "vanilla" client server calls, the server can now connect to selected drivers
    while the other drivers still running.
**Fix c:**
    Replace the 2.7 oriented libraries with 3.x native libraries, websockets and aiohttp.

**What is changing:**

The admin needs to actively select the drivers to run the match instead of waiting for two students to connect.
All students must run there drivers before the match begin.

**Running locally:**

``` bash
# Run the client using an example driver (default port 8081, use --port to change)
./rose-client --driver examples/random-driver.py

# On a different terminal (tmux session)
./rose-server

# Connect the server to the client
./rose-admin http://127.0.0.1:8880 set-drivers http://127.0.0.1:8081

# Start a match
./rose-admin http://127.0.0.1:8880 start
```


yaacov:
@nirs can you take a look ?
@bennypowers hi, can you take a look at the js parts ?
@slaviered hi, this should not change the docs and classroom documentation, if it does, I will fix them.

rose/server/web/index.html
```diff
@@ -9,11 +9,20 @@
 
   <body>
     <div id="content">
+      <a href="settings.html" id="settings-link">Settings</a>
```
bennypowers, line 12: you might want to make this `settings/` and rename settings.html to settings/index.html

rose/server/web/index.html
```diff
@@ -9,11 +9,20 @@
...
+      <a href="#" id="info-btn">Info</a>
+
+      <div id="info-panel" class="hidden">
```
bennypowers, line 15: class=hidden vs hidden vs use a pre-built custom-element with aria announce stuff ready to go

rose/server/web/index.html
```diff
@@ -9,11 +9,20 @@
...
+          <span id="info-text"></span>
+      </div>
+
       <form id="control">
 
-        <button id="start" disabled>Start</button>
+        <button id="run" disabled>Run</button>
         <button id="stop"  disabled>Stop</button>
         <button id="music_ctl">Music</button>
+        <button id="reset">Reset</button>
```
bennypowers, line 25:
```suggestion
        <button id="reset" type="reset">Reset</button>
```

rose/server/web/settings.css
```diff
@@ -0,0 +1,84 @@
+body.dark-theme {
+    background-color: #2c3e50;
```
bennypowers, line 2: consider using red hat design tokens

https://ux.redhat.com/tokens

rose/server/web/settings.html
```diff
@@ -0,0 +1,76 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Set Drivers</title>
+    <link rel="stylesheet" href="settings.css">
+    <script>
```
bennypowers, line 8: prefer not to block rendering: make this a module, and wrap the functionality in a custom element

rose/server/web/settings.html
```diff
@@ -0,0 +1,76 @@
...
+            window.location.href = "index.html";
+        }
+
+        function sendDrivers(event) {
+            // Prevent the form from submitting in the traditional way
+            event.preventDefault();
+
+            const driver1 = document.getElementById('driver1').value;
+            const driver2 = document.getElementById('driver2').value;
+            const formattedData = `drivers=${driver1},${driver2}`;
```
bennypowers, line 37: use [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) constructor and [`SearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams) object

rose/server/web/settings.html
```diff
@@ -0,0 +1,76 @@
...
+            // Send the formatted data to the server using an AJAX POST request
+            fetch(`/admin?${formattedData}`, {
+                method: 'POST'
+            })
+            .then(response => {
+                if(response.ok) {
+                    // If the POST request was successful, redirect to index.html
+                    window.location.href = "index.html";
+                } else {
+                    console.error('Error:', response.statusText);
```
bennypowers, line 48: prefer to parse and display errors on screen

rose/server/web/settings.html
```diff
@@ -0,0 +1,76 @@
...
+                }
+            })
+            .catch(error => {
+                console.error('Error:', error);
+            });
+        }
+
+        document.addEventListener('DOMContentLoaded', loadDrivers);
```
bennypowers, line 56: prefer a module, which is always defered

rose/server/web/settings.html
```diff
@@ -0,0 +1,76 @@
...
+    </script>
+</head>
+<body class="dark-theme">
```
bennypowers, line 59: use `prefers-color-scheme`

rose/server/web/settings.html
```diff
@@ -0,0 +1,76 @@
...
+    <h2 class="heading">Set Game Drivers</h2>
+    <form onsubmit="sendDrivers(event)" class="form">
+        <div class="form-group">
+            <label for="driver1" class="label">Driver 1:</label>
+            <input type="text" id="driver1" name="driver1" placeholder="http://127.0.0.1:8081" class="input-field">
```
bennypowers, line 64: usability: due to the nature of this project, it might not be possible to avoid entering loopback IP addresses, but if we can, we should

rose/server/web/settings.html
```diff
@@ -0,0 +1,76 @@
...
+        </div>
+        <div class="form-group">
+            <label for="driver2" class="label">Driver 2:</label>
+            <input type="text" id="driver2" name="driver2" class="input-field">
+        </div>
+        <div class="form-group">
+            <button type="submit" class="submit-button">Send</button>
```
bennypowers, line 71: don't need the type here. or really the class

rose/server/web/settings.html
```diff
@@ -0,0 +1,76 @@
...
+            <button type="button" class="cancel-button" onclick="cancelSetting()">Cancel</button>
```
bennypowers, line 72: don't need the type here

and prefer to attach event listeners in the module. consider wrapping the form in a custom element that handles the behaviour

```html
<rose-drivers>
  <form>...</form>
</rose-drivers>
```

nirs:
Cool demo!

But about 20x times too big to review. Can you break up to small easy to review commits?

In general the idea that the server connects to the drivers looks strange and harder to use.

- you need to run the server after the clients
- you need to know the clients address to run the server
- you need to restart the server for every game
- the server need to handle client disconnection - previously the client could reconnect

The current model when the clients connects to the server is better. I don't see why we need
to change it.

Regarding the problems:

> a. We can not create a k8s Pod for each driver, and let them run forever.
When a driver starts it immediately try to connect to the server and blocks other drivers from running.

Why do you want to run the client forever? The client is something you run temporarily when testing or during a game.

> b. We use python RPC protocol to comunicate between the driver and the server, this makes it hard to write drivers in compiled languages, making us relay on .pyc file to hide implementation code.

This is incorrect. We use simple line based JSON lines RPC protocol that is easy to write
any any languages and is not related to python in any way.

The protocol for for communicating with the server is not related to the the driver pyc file in any way. The dependency is the python client which can be replaced with a client written in any other language.

> c. We use 2.7 oriented libraries: twisted, autobahn and pytest.

These libraries are not python 2.7 oriented, they just mature libraries that were popular
years before python 3 was started.

Twisted is awesome because it let you run multiple services clients in the same process. I don't think there is anything replacing it. For example, if you want to add an IRC clinet reporting game results to IRC channel, this is really easy in twisted.

autobahn is web socket library with twisted integartion. websoket is the best way to commmunicate with the game UI, allowing real time updates. Why replace it?

pytest is the best testing library for python (if not the best testing library for anything).
I'm not sure what is the issue with it.

Since we used twisted, it was very easy to add xmlrpc server for the admin tool, but we
should realy get rid of the xmlrpc code since simple http (already used by the UI) is
good enough for the admin tool, or maybe we can just drop the admin tool - since the
web UI was added, there is no need for the admin tool.

It will be more productive to open issues for any problem and discuss them.

yaacov:
@nirs thanks

This PR is to make the Game more Openshift ready, when I run pods in Openshift, I want to deploy my pods, and let them run.

> Why do you want to run the client forever? The client is something you run temporarily when testing or during a game.

"As an instructor I want to ask the students to deploy their drivers to the cluster, and then run matches using an already running server and drivers"

removing  the xmlrpc connection between the driver and server makes it easy to run the game-engine (named "server") and the driver (named "client") as Pods that live as good citizens of the cluster.

> you need to run the server after the clients

you don't need to re-run the server, the idea is that the server is running all the time.
you only need to choose the drivers for the next match, and you do it while the server is running.

> you need to know the clients address to run the server

this is easy when running on a cluster:

``` bash
oc get services -n students
```

> you need to restart the server for every game

the server is good citizen of the cluster, it is deployed once and run forever,
the only caveat is that the game engine "server" has a state while running a game, so you can only have one "server" serving a set of games.

you can run several game-engines "servers" but they will need to have a URL per server and can't run using a load balancer, the drivers on the other hand are stateless and can run in deployments of more then one.

> the server need to handle client disconnection - previously the client could reconnect

the game-engine (named "server" because it is currently called this way) is a client or the driver (named "client" because it is currently called this way) server, and does not care  if the driver is connected or not, it just pool it, it can get an answer of miss it, nothing will happen if no answer is given except the the car will continue forward and the user will get a notice.

yaacov:
> This is incorrect. We use simple line based JSON lines RPC protocol that is easy to write
any any languages and is not related to python in any way.

> The protocol for for communicating with the server is not related to the the driver pyc file in any way. The dependency is the python client which can be replaced with a client written in any other language.

@nirs ok, so it uses twisted protocol instead of xmlrpc ...
https://twisted.org/documents/8.1.0/api/twisted.protocols.basic.LineReceiver.html

not sure if it's that easy to implement in other languages, but I didn't try :-)
I still think that simple HTTP client server is more clean and clear, and easy to implement in "any language" (tm)

yaacov:
> Can you break up to small easy to review commits?

Yes, I will see how to make it more clear :+1:

yaacov:
Twisted vs asyncio

asyncio is a part of the standard library. no need to install it, it's a nice plus for using it.
asyncio: Uses async/await syntax, this is debatable, but I think that most users like it, this pattern is very common in other langauges too, so it's also a plus.
asyncio: works nice with aiohttp, while twisted works nice with autobahn ... so it's a tie here :-)

To sum things up, Twisted is awesome, but asyncio is integrated with python and is "the new cool thing"
for me it's 2 plus for asyncio and one tie :-)

yaacov:
pytest vs unittest

unittest: Part of the Python standard library. You don't need to install anything extra to use it.
unittest: Doesn't support parameterized testing out of the box. You'd need to use additional tools or libraries

so it's one plus for unittest and one for pytest :-)
but the fact it's part of Python makes me think it's nicer to use.

yaacov:
> asyncio does not replace twisted - it is just the core of twisted, and I think that twisted
can use the asyncio event loop if you want.

For basic HTTP client server communication, you don't need twisted, asyncio is doing all that is needed.
Anything fancier that require twisted will make it harder to implement using tools and languages that don't have twisted.

> I cannot think about more unrelated topic to making the container more friendly :-)

ok :-)

yaacov:
@nirs hi,

a. reverted ci to use pytest, still need to translate the tests.
b. started splitting the PR, now it's split between metadata commits and code commits, still need to split the code commits further.

rose/server/web/settings.html
yaacov: using web component "settings-form"

rose/server/web/settings.html
yaacov: added a type to prevent form event trigger, can I avoid it ?

rose/server/web/settings.html
yaacov: changed to a Kubernetes like URL
"http://driver.student.svc.cluster.local:8081"

rose/server/web/settings.html
yaacov: moved to using module

rose/server/web/settings.html
yaacov: removed the console.errors

rose/server/web/settings.html
yaacov: yeap, fixed :-)

rose/server/web/settings.html
yaacov: moved to module

rose/server/web/settings.html
yaacov: ahh, sorry, it now a regular button ... this one do need the type :-)

bennypowers: if you need to prevent default, do it in the custom element class, otherwise havea route on the http server that accepts the http formdata, for progressive enhancement

rose/server/web/index.html
yaacov: moved to '/settings/index.html'
note that the python server we are using can't automatically show `index.html` files when the request is `/` it show 404

rose/server/web/index.html
yaacov: ARIA is a little out of scope here :-)
we may implement when we move to web components

rose/server/web/index.html
yaacov: @bennypowers i currently put the web component in `settings-form` file in the same dir, is there a convention for how to name web components ? where to put tehm ?

rose/server/web/settings/settings-form.js
```diff
@@ -0,0 +1,82 @@
...
+
+    connectedCallback() {
+        this.render();
+        this.loadDrivers();
+        this.shadowRoot.querySelector('#submit-button').addEventListener('click', this.sendDrivers.bind(this));
+        this.shadowRoot.querySelector('#cancel-button').addEventListener('click', this.cancelSetting.bind(this));
+    }
+
+    render() {
+        this.shadowRoot.innerHTML = `
```
bennypowers: Interesting take. I would have thought to not use shadow DOM in this case and just this.querySelector, but this is good too

You may choose to use a template element is a static init block, but this isn't reusable so the gains are few

If you'll stick with shadow DOM, use a constructible style sheet and adopted stylesheets instead of a link element

rose/server/web/index.html
bennypowers: Usually, name the module the same as the tag name

rose/server/web/index.html
bennypowers: While aria is not necessary, the experience must be available to screen reader users

rose/server/web/settings/settings-form.js
yaacov: @bennypowers that was a nice learning experience :-)
not using template, but separated the template and css from the main js component, WDYT ?

rose/server/web/settings/index.html
```diff
@@ -0,0 +1,16 @@
...
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Set Drivers</title>
+    <link rel="stylesheet" href="./index.css">
+</head>
+<body class="dark-theme">
+    <h2 class="heading">Set Game Drivers</h2>
+    <settings-form class="form" />
```
bennypowers: custom-elements can't be void
```suggestion
    <settings-form class="form"></settings-form>
```

bennypowers (changes requested):

rose/server/web/settings/index.html
yaacov: wow, fixed, btw/ why did it work ?

yaacov: I see it was a hot topic :fire: 
https://github.com/WICG/webcomponents/issues/624

yaacov:
Closing in favor of specific repository per game container:

| component | repo |
|----|---|
Game engine	| https://github.com/RedHat-Israel/rose-game-engine |
Game web based user interface	| https://github.com/RedHat-Israel/rose-game-web-ui |
Game car driving module	| https://github.com/RedHat-Israel/rose-game-ai |

I also added a reference container for running machine learning driver:
https://github.com/RedHat-Israel/rose-game-ai-reference

to run on all components on a cluster:

`kubectl apply -f https://raw.githubusercontent.com/RedHat-Israel/rose-game-engine/main/rose-game.yaml
`

to run locally:
```
podman run --rm --network host -it quay.io/rose/rose-game-engine:latest
podman run --rm --network host -it quay.io/rose/rose-game-web-ui:latest
```

and for the driver, create `mydriver.py` file locally, and run:
```
podman run --rm --network host -it \
  -v $(pwd)/mydriver.py:/mydriver.py:z \
  -e DRIVER /mydriver.py \
  quay.io/rose/rose-game-ai:latest
```

PR #487, opened 2024-07-14
Adjust Dockerfile to new version of pipenv

reihl:
pipenv has been change and now you need to use pipenv requirements to create requirements.txt file

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,9 +14,11 @@ COPY Pipfile ./Pipfile
 # Generate requirements.txt file from Pipfile
 RUN if [ -z ${DEV} ]; \
     then \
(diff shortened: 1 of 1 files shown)


sleviim:
@nirs Hi Nir,
Can we merge this one?
Otherwise, students will need to run it manually

nirs:
Getting rid of pipenv is better, but if this works it is an easier change. This project does not need pipenv locking. pip install with requirements.txt file is better and much faster.

yaacov:
Hi guys, we have a container friendly version in, no pipenv is used:
 https://github.com/RedHat-Israel/rose-game-engine

